PR #35, opened 2023-03-31
Added support for docker ARM, M1 MACS

c0d3G33k:


.github/workflows/container.yml
```diff
@@ -44,7 +44,7 @@ jobs:
 
     - name: Build and Push Container Image
       run: |
-        docker buildx build --push --platform linux/amd64 \
+        docker buildx build --push --platform linux/amd64, linux/arm64 \
```
abhisek, line 47: I think we need to use `--platform linux/amd64 --platform linux/arm64`

Without that it will error out

```
docker buildx build --platform linux/amd64, linux/arm64 -t safedep/vet .
ERROR: "docker buildx build" requires exactly 1 argument.
See 'docker buildx build --help'.
```

abhisek (changes requested):

abhisek (approved):

PR #45, opened 2023-04-06
adding homebrew-vet tap

madhuakula:


.goreleaser.yaml
```diff
@@ -7,13 +7,30 @@ before:
...
+  - replace: true
+
+brews:
+  - 
+    name: vet
+    homepage: https://safedep.io
+    description: "SafeDep vet is a tool for identifying open source software supply chain risks"
+    tap:
+      owner: safedep
+      name: homebrew-vet
```
abhisek, line 32: We need to create this repository and configure this job with a PAT that has access to push to the repo.

Also is it possible to create a common brew repo, something like `safedep/homebrew` and push `vet` formula there along with any other tool that we develop in future?

madhuakula: Fixed this to be `homebrew-tap` as the releaser looks for `homebrew-*`. So now we can leverage this for all the SafeDep tooling 👍

abhisek (approved):

Issue #55, opened 2023-04-09
Fix Linter Issues and Enable `golint` Guard Rail


abhisek:
We currently have quite a few linter issues.go

```
✗ golint ./...  | wc -l
124
```

We need to fix them and introduce a linter guard rail in 
https://github.com/safedep/vet/blob/main/.github/workflows/ci.yml

The guard rail can be as simple as integrating
https://github.com/golangci/golangci-lint-action

tarunsamanta2k20:
I would like to work on this issue.

PR #58, opened 2023-04-09
Added linter support on ci

tarunsamanta2k20:
Ask me for changes if any. 

This commit is only for testing.

Please run all check once. Let's see what happens.

#55 
@tarunsamanta2k20


.github/workflows/ci.yml
```diff
@@ -25,3 +25,10 @@ jobs:
           go mod tidy
           go build
           go test -v ./...
+          
+      - name: golangci-lint
```
abhisek, line 29: I would suggest that we run the linter checks before the building and running tests. The reason is, linter checks are relatively fast compared to build and test. So we will consume less CI minutes if we just fail fast with linter checks before running build and test.

abhisek: Also can you verify if the `golangci/golangci-lint-action@v3` lints only the PR diffs or entire code base? It will be great if we can lint only the changed files to start with. This is because we will probably have quite a lot of linter failures in existing code (tech debt) which we need to fix incrementally.

abhisek (changes requested):

abhisek:
@tarunsamanta2k20 The CI linter check is failing. I think the problem is similar to
https://github.com/golangci/golangci-lint-action/issues/434

tarunsamanta2k20:
This PR I kept as draft PR. I will work on it.

abhisek:
@tarunsamanta2k20 This is implemented.

Issue #54, opened 2023-04-09
Show Ecosystem Name in Summary Report


abhisek:
The summary reporter presents a table of packages that are recommended for upgrade. It does not show the ecosystem of the package and only shows the name and version. While this is fine for a scan where only a single lockfile was scanned, this is a problem where multiple lockfiles were scanned with different ecosystem

![Screenshot-2023-04-09-170817](https://user-images.githubusercontent.com/31844/230770338-f1467712-bd2e-41ec-bced-2e0f633f2d2d.png)

To start improving the UX, we should start by showing the Ecosystem name in the report report table.

tarunsamanta2k20:
I'd like to work on this issue.

tarunsamanta2k20:
I would like to work on this issue.

PR #59, opened 2023-04-09
Added feature for showing Ecosystem name in the table along with othe…

tarunsamanta2k20:
Issue description:

The summary reporter presents a table of packages that are recommended for upgrade. It does not show the ecosystem of the package and only shows the name and version. While this is fine for a scan where only a single lockfile was scanned, this is a problem where multiple lockfiles were scanned with different ecosystem

Screenshot-2023-04-09-170817

To start improving the UX, we should start by showing the Ecosystem name in the report report table.

Let me know if there are any changes instead of closing this PR.

What is drift I don't know I kept as it is.

#54 
image

### DIFF
--- a/pkg/reporter/summary.go
+++ b/pkg/reporter/summary.go
@@ -234,7 +234,6 @@ func (r *summaryReporter) renderRemediationAdvice() {
 
 	fmt.Println(text.Bold.Sprint("Consider upgrading the following libraries for maximum impact:"))
 	fmt.Println()
-	// fmt.Println(sortedPackages)
 	tbl := table.NewWriter()
 	tbl.SetOutputMirror(os.Stdout)
 	tbl.SetStyle(table.StyleLight)
--- a/pkg/reporter/summary.go
+++ b/pkg/reporter/summary.go
@@ -251,6 +251,7 @@ func (r *summaryReporter) renderRemediationAdvice() {
 			r.packageNameForRemediationAdvice(sp.pkg),
 			utils.SafelyGetValue(insight.PackageCurrentVersion),
 			sp.score,
+			r.packageNameForRemediationAdvice(sp.ecosystem),
 		})
 
 		tagText := ""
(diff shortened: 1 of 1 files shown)


pkg/reporter/summary.go
```diff
@@ -248,9 +248,11 @@ func (r *summaryReporter) renderRemediationAdvice() {
 		insight := utils.SafelyGetValue(sp.pkg.Insights)
 
 		tbl.AppendRow(table.Row{
+			sp.ecosystem,
```
abhisek, line 251: Looks good. You need to add an appropriate header row as well for this additional column. Have a look at https://github.com/safedep/vet/blob/main/pkg/reporter/summary.go#L242

abhisek (changes requested):

pkg/reporter/summary.go
```diff
@@ -248,9 +248,11 @@ func (r *summaryReporter) renderRemediationAdvice() {
...
 			r.packageNameForRemediationAdvice(sp.pkg),
 			utils.SafelyGetValue(insight.PackageCurrentVersion),
 			sp.score,
+
```
abhisek, line 255: Do we really need an empty line here?

abhisek (changes requested):

pkg/reporter/summary.go
tarunsamanta2k20: So you are saying I have to add a new column as well. Right?

abhisek:
@tarunsamanta2k20 Please try to build your branch. If you have Go tool chain setup, you can build `vet` just with:

```
go build
```

pkg/reporter/summary.go
abhisek: Yes. You need to add a new column for the header row as well.

tarunsamanta2k20:
@abhisek which ecosystem variable should i use? I mean it is defied in lots of files.

tarunsamanta2k20:
@abhisek 

image

abhisek:
> @abhisek
> 
> image

@tarunsamanta2k20 Looks great. Just a small request. Can we have `Ecosystem` as the first column? The justification for this is so that `Package` column which includes both name and version is followed by `Update Version` i.e. the latest version available for the package. 

The table will be more intuitive if we have the the version to update just next to the package.

abhisek (approved):

tarunsamanta2k20:
@abhisek please add me as a member of your organization. I wanna contribute more in tte future.

PR #69, opened 2023-04-12
Adds CSV Reporter for Output : Issue :#6

shivamsk:
PR to **Support CSV Output for Filter Results #6**

[https](url)://github.com/safedep/vet/issues/6

### DIFF
--- a/pkg/reporter/csv.go
+++ b/pkg/reporter/csv.go
@@ -1,0 +1,94 @@
+package reporter
+
+import (
+	"encoding/csv"
+	"fmt"
+	"os"
+
+	"github.com/safedep/vet/pkg/analyzer"
+	"github.com/safedep/vet/pkg/common/logger"
+	"github.com/safedep/vet/pkg/models"
+	"github.com/safedep/vet/pkg/policy"
+	"github.com/safedep/vet/pkg/readers"
+)
+
+
+
+type csvReporter struct {
+	csvRecords []CsvRecord
+}
+
+func NewCsvReporter() (Reporter, error) {
+	return &csvReporter{}, nil
+}
+
+func (r *csvReporter) Name() string {
+	return "Csv Report Generator"
+}
+
+func (r *csvReporter) AddManifest(manifest *models.PackageManifest) {
+	r.csvRecords = make([]CsvRecord,0)
+	readers.NewManifestModelReader(manifest).EnumPackages(func(pkg *models.Package) error {
+		r.csvRecords = append(r.csvRecords,r.createCsvRecord(pkg))
+		return nil
+	})
+}
+
+func (r *csvReporter) AddAnalyzerEvent(event *analyzer.AnalyzerEvent) {}
+
+func (r *csvReporter) AddPolicyEvent(event *policy.PolicyEvent) {}
+
+func (r *csvReporter) Finish() error {
+
+	csvResponse := r.generateCsv(r.csvRecords)
+
+	// Error case
+	if(csvResponse != nil){
+		return csvResponse
+	}
+	return nil
+}
+
+type CsvRecord struct {
+	packageName string
+	updateTo    string
+}
+
+func (r *csvReporter) createCsvRecord(pkg *models.Package) *CsvRecord {
+	return &CsvRecord{
+		packageName: r.packageNameForRemediationAdvice(pkg),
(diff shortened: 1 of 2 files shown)


pkg/reporter/csv.go
```diff
@@ -0,0 +1,92 @@
...
+	packageName string
+	updateTo    string
+}
+
+func NewCsvReporter() (Reporter, error) {
+	return &csvReporter{}, nil
+}
+
+func (r *csvReporter) Name() string {
+	return "Csv Report Generator"
```
abhisek, line 29: nit: Can we please use `CSV` instead of `Csv` in name?

pkg/reporter/csv.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func (r *csvReporter) AddManifest(manifest *models.PackageManifest) {
+	r.csvRecords = make([]CsvRecord,0)
+	readers.NewManifestModelReader(manifest).EnumPackages(func(pkg *models.Package) error {
```
abhisek, line 34: What are we trying to achieve here? Apologies for not writing the requirement clearly in the Github issue.

I am not sure if we should add every package into the CSV output because it will have too much data (aka. noise). How about we only add packages that match a filter into CSV output? For example:

```
vet scan -D /path --filter 'vulns.critical.exists(p, true)' --report-csv=/tmp/critical-vuln-packages.csv
```

See example here
https://github.com/safedep/vet/blob/main/pkg/reporter/markdown.go#L77

Thoughts?

scan.go
```diff
@@ -82,6 +83,10 @@ func newScanCommand() *cobra.Command {
 		"Print a report to the console")
 	cmd.Flags().BoolVarP(&summaryReport, "report-summary", "", true,
 		"Print a summary report with actionable advice")
+	cmd.Flags().BoolVarP(&csvReport, "report-csv", "", false,
```
abhisek, line 86: This should be a `StrinVarP` no? We need a file to write the CSV report

scan.go
```diff
@@ -82,6 +83,10 @@ func newScanCommand() *cobra.Command {
...
+		"Generate a csv report")
+	cmd.Flags().BoolVarP(&csvReport, "report-csv", "", false,
```
abhisek, line 88: Why are we duplicating this?

abhisek (changes requested):

abhisek:
@shivamsk Please ensure your branch is building. It seems there are syntax error in your branch.

pkg/reporter/csv.go
shivamsk: @abhisek Will change it to Path. 
So, it should be similar to the markdown report built on top of the summary reporter. Please correct me if I am wrong. 
What all fields should be present in CSV file headers?

abhisek: You can consider building the CSV reporter on top of summary reporter. I haven't considered the possibility but seems like it is an option because summary reporter already handles recording of events raised by Analyzers (Filter). 

I would consider the following fields in CSV to start with:

```
ecosystem, manifest path, packageName, packageVersion, filter violation reason (available from the event)
```

@c0d3G33k Any suggestion on what all fields will be useful in CSV report?

shivamsk: @abhisek Got it. 
I am assuming we generate the CSV records only for violations as only they will have violation reason listed. 
Is my assumption correct ? 
https://github.com/safedep/vet/blob/main/pkg/reporter/markdown.go#L134

shivamsk: @abhisek Pushed the changes. Please review

pkg/reporter/csv.go
```diff
@@ -0,0 +1,130 @@
...
+		violations:      make(map[string]*analyzer.AnalyzerEvent),
+	}, nil
+}
+
+func (r *csvReporter) Name() string {
+	return "CSV Report Generator"
+}
+
+func (r *csvReporter) AddManifest(manifest *models.PackageManifest) {
+	r.summaryReporter.AddManifest(manifest)
```
abhisek, line 48: Do we really need the summary reporter? I do not see any reference to `summaryReporter` while generating CSV. We can just ignore the manifest here I think.

abhisek (changes requested):

pkg/reporter/csv.go
shivamsk: @abhisek Removed it.

Issue #118, opened 2023-08-18
Vet Crash on one of the SBOM generate from Github


jchauhan:
Attached screenshot for the reference

![bug_vet_sbom](https://github.com/safedep/vet/assets/74857/1aceaded-e204-4a34-a131-0ec5cddc1e39)

PR #119, opened 2023-08-18
Fix: Vet Crash on one of the SBOM generate from Github #118

jchauhan:


pkg/parser/cyclonedx_sbom.go
```diff
@@ -14,25 +14,34 @@ import (
...
 		}
 	}
 
-	for _, comp := range *bom.Components {
-		if d, err := convertSbomComponent2LPD(&comp); err != nil {
-			logger.Warnf("Failed converting sbom to lockfile component: %v", err)
-		} else {
-			details = append(details, *d)
+	// Components is a pointer array and it can be empty
+	if bom.Components != nil {
```
abhisek, line 33: You can consider using https://github.com/safedep/dry/blob/main/utils/strings.go#L35

```suggestion
components = utils.SafelyGetValue(bom.Components)
for _, comp := range components {
...
}
```

PR #122, opened 2023-08-24
feat: Add Support for Github Connect for Private Repository Scanning

abhisek:


abhisek:
Linter failure is FP. We are falling through the reader init and then checking error in a single place.

PR #123, opened 2023-08-25
Fixed Issue with NPM sbom, fixed issues with cyclonedx parser, and refactored code

jchauhan:


pkg/common/utils/sbom_utils/sbom.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"github.com/google/osv-scanner/pkg/lockfile"
+	"github.com/package-url/packageurl-go"
+)
+
+// ParsePurlType parses a package URL (purl) type and returns the corresponding lockfile.Ecosystem.
+// It also returns a boolean indicating whether the provided purl type is recognized.
+// The recognized purl types are mapped to their corresponding lockfile ecosystems.
+func ParsePurlType(purl_type string) (lockfile.Ecosystem, bool) {
```
abhisek, line 11: Can we rename this to something very explicit such as `ParsePurlTypeToLockfile` and return `error` in case of failure instead of `bool`. Error helps us carry the cause of failure along the control flow and is easier to debug with logs. Example error when known type is not available:

```
fmt.Errorf("no known package url type found for: %s", purl_type)
```

pkg/common/utils/sbom_utils/sbom_test.go
```diff
@@ -0,0 +1,37 @@
+package sbom_utils
```
abhisek, line 1: As per go convention, we should use a package name such as `sbomutils`

> By convention, packages are given lower case, single-word names; there should be no need for underscores or mixedCaps

https://go.dev/doc/effective_go

pkg/common/utils/sbom_utils/sbom.go
```diff
@@ -0,0 +1,29 @@
+package sbom_utils
```
abhisek, line 1: As per go convention, we should use a package name such as `sbomutils`

> By convention, packages are given lower case, single-word names; there should be no need for underscores or mixedCaps

https://go.dev/doc/effective_go

pkg/parser/custom/cyclonedx_sbom/cyclonedx.go
```diff
@@ -1,18 +1,17 @@
-package parser
+package cyclonedx_sbom
```
abhisek, line 1: As per go convention, we should use a package name such as `sbomutils`

> By convention, packages are given lower case, single-word names; there should be no need for underscores or mixedCaps

https://go.dev/doc/effective_go

pkg/parser/custom/packagefile/packagefile.go
```diff
@@ -1,8 +1,13 @@
 package packagefile
```
abhisek, line 1: This is good. Package name should be like this

pkg/parser/custom/packagefile/packagefile.go
```diff
@@ -38,14 +44,51 @@ type PackageDetails struct {
 	CycloneDxRef *cdx.Component     `json:"cylcone_dx_ref,omitempty"`
```
abhisek, line 44: `PackageDetails` is referring to `CycloneDX` which makes me think that this data structure probably belongs within the boundary of `cyclonedxsbom` package and not in a package of its own. What do you think?

abhisek (changes requested):

pkg/parser/custom/packagefile/packagefile.go
jchauhan: No,  it is in the boundary of Custom parsers, and specially spdx and cyclonedx both. In my view, we should have built a wrapper over lockfile package details as we need to maintain more state than that is provided by it. PackageDetails have enough context including source of conversion so that we can convert it to more richer reports later.

abhisek (approved):

Issue #163, opened 2023-11-30
Vet is not able to detect setup.py manifest given a directory as scan input 


jchauhan:
## Current Result

setup.py manifest is not detected

```bash 
vet  scan -D ../**/tmp/***/ 

 .----------------.  .----------------.  .----------------.
| .--------------. || .--------------. || .--------------. |
| | ____   ____  | || |  _________   | || |  _________   | |
| ||_  _| |_  _| | || | |_   ___  |  | || | |  _   _  |  | |
| |  \ \   / /   | || |   | |_  \_|  | || | |_/ | | \_|  | |
| |   \ \ / /    | || |   |  _|  _   | || |     | |      | |
| |    \ ' /     | || |  _| |___/ |  | || |    _| |_     | |
| |     \_/      | || | |_________|  | || |   |_____|    | |
| |              | || |              | || |              | |
| '--------------' || '--------------' || '--------------' |
 '----------------'  '----------------'  '----------------'

Running in Community Mode
Scanning packages    ... done! [0 in 2ms]
Scanning manifests   ... done! [0 in 2ms]
  **   Summary of Findings 

  ** 0 critical, 0 high and 0 other vulnerabilities were identified

  ** 0 potentially unpopular library identified as direct dependency

  ** 0 libraries are out of date with major version drift in direct dependencies

  ** across 0 libraries in 1 manifest(s)

Consider upgrading the following libraries for maximum impact:

┌───────────┬─────────┬───────────┬────────┐
│ ECOSYSTEM │ PACKAGE │ UPDATE TO │ IMPACT │
├───────────┼─────────┼───────────┼────────┤
└───────────┴─────────┴───────────┴────────┘

Run with `vet --filter="..."` for custom filters to identify risky libraries
For more details https://github.com/safedep/vet
```

Expected Result

```bash 
❯ ./vet  scan -D ../***/tmp/***/

 .----------------.  .----------------.  .----------------.
| .--------------. || .--------------. || .--------------. |
| | ____   ____  | || |  _________   | || |  _________   | |
| ||_  _| |_  _| | || | |_   ___  |  | || | |  _   _  |  | |
| |  \ \   / /   | || |   | |_  \_|  | || | |_/ | | \_|  | |
| |   \ \ / /    | || |   |  _|  _   | || |     | |      | |
| |    \ ' /     | || |  _| |___/ |  | || |    _| |_     | |
| |     \_/      | || | |_________|  | || |   |_____|    | |
| |              | || |              | || |              | |
[...]

PR #164, opened 2023-11-30
Fix for the issue #163 - Vet is not able to detect setup.py manifest …

jchauhan:
…given a directory as scan input

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -2,6 +2,7 @@ package parser
 
 import (
 	"fmt"
+	"path"
 
 	"github.com/google/osv-scanner/pkg/lockfile"
 	"github.com/safedep/vet/pkg/common/logger"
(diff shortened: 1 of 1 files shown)


pkg/parser/parser.go
```diff
@@ -87,18 +88,31 @@ func FindParser(lockfilePath, lockfileAs string) (Parser, error) {
 		}
 	}
 
+	pw, _ := findExperimentalParser(lockfilePath, lockfileAs)
```
abhisek, line 91: I don't think we should include experimental parsers automatically while scanning a directory because they primarily consist of parsers that does not have a standardised input file names. For example, SPDX / CycloneDX SBOM can be in any JSON file within the directory. `setup.py` can be an application code and not a manifest in which case there is a possibility that our parser will fail (which is fine) or generate false positive (which we should avoid).

This is the reason why we have separate list of experimental parsers that can be used explicitly using `--lockfile-as` and not selected automatically during a directory scan.

jchauhan: No Problem, i will switch to another tool that will support setup.py because I need to parse it. Most of the python libraries will have setup.py.

PR #255, opened 2024-10-17
Chore/goreleaser upgrade

insaaniManav:
Upgraded goreleaser to v2

### DIFF
--- a/.github/workflows/goreleaser.yml
+++ b/.github/workflows/goreleaser.yml
@@ -78,7 +78,7 @@ jobs:
         uses: goreleaser/goreleaser-action@286f3b13b1b49da4ac219696163fb8c1c93e1200 # v6.0.0
         with:
           distribution: goreleaser
-          version: '~> v2'
+          version: 'v2'
           args: release --clean --skip publish --snapshot
         env:
           GITHUB_TOKEN: ${{ secrets.GORELEASER_GITHUB_TOKEN }}
--- a/.github/workflows/goreleaser.yml
+++ b/.github/workflows/goreleaser.yml
@@ -77,7 +77,7 @@ jobs:
         with:
           distribution: goreleaser
           version: 'v2'
-          args: release --clean --skip publish
+          args: release --clean --skip publish --snapshot
         env:
           GITHUB_TOKEN: ${{ secrets.GORELEASER_GITHUB_TOKEN }}
 
--- a/.github/workflows/goreleaser.yml
+++ b/.github/workflows/goreleaser.yml
@@ -2,6 +2,8 @@ name: Release Automation
 
 on:
   push:
+    tags:
+      - "*" # triggers only if push new tag version, like `0.8.4` or else
 
 concurrency: ci-release-automation
 
@@ -74,7 +76,7 @@ jobs:
         with:
           distribution: goreleaser
           version: '~> v2'
-          args: release --clean --skip publish --snapshot
+          args: release --clean
         env:
           GITHUB_TOKEN: ${{ secrets.GORELEASER_GITHUB_TOKEN }}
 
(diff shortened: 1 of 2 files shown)


.github/workflows/goreleaser.yml
```diff
@@ -75,9 +75,10 @@ jobs:
 
       - name: Run GoReleaser
         id: run-goreleaser
-        uses: goreleaser/goreleaser-action@8f67e590f2d095516493f017008adc464e63adb1 # v4.1.0
+        uses: goreleaser/goreleaser-action@v6 # v4.1.0
```
abhisek, line 78: Can we please pin the version to the SHA? This is a safeguard against malicious actions because the tag `v6` may be updated to point to a malicious version of the action. Pinning to SHA sum ensures immutability of the action.

.github/workflows/goreleaser.yml
```diff
@@ -75,9 +75,10 @@ jobs:
...
         with:
-          version: v1.26.2
+          distribution: goreleaser
+          version: 'v2'
```
abhisek, line 81: Should we ping to a specific version instead of the major version `v2` to avoid surprises with automatic version changes? Open to discussion / thoughts here.

abhisek (changes requested):

abhisek:
@insaaniManav Thanks for the PR. Are you able to run the `goreleaser` binary with this config locally? I don't have a good way to test this without triggering a release.

.goreleaser.yaml
```diff
@@ -10,31 +11,15 @@ env:
...
-      - CC=o64-clang
-      - CXX=o64-clang++
-
-  - id: darwin-arm64
-    binary: vet
-    goos:
-      - darwin
-    goarch:
-      - arm64
+  - id: darwin
```
abhisek, line 20: Nice. I like this simplification.

insaaniManav:
> @insaaniManav Thanks for the PR. Are you able to run the `goreleaser` binary with this config locally? I don't have a good way to test this without triggering a release.

yup here is a dry run https://github.com/insaaniManav/vet/actions/runs/11379032341 - will make the requested changes - trigger another dry run and then get back here

insaaniManav:
@abhisek I have pinned the commit hash in the run , also I have set the version to '~> v2' which is the official recommend thing , here is the successful workflow https://github.com/insaaniManav/vet/actions/runs/11381193881

.github/workflows/goreleaser.yml
```diff
@@ -100,5 +98,4 @@ jobs:
     with:
       base64-subjects: "${{ needs.goreleaser.outputs.hashes }}"
       upload-assets: true
-      private-repository: true
-
+      private-repository: true
```
abhisek, line 101: Do we need this? vet is a public repository. What's the benefit of keeping this private?

.github/workflows/goreleaser.yml
```diff
@@ -62,22 +61,21 @@ jobs:
...
           SDK_VERSION=$OSX_CROSS_MACOS_SDK_VERSION UNATTENDED=yes ./build.sh
-
       - name: Add OSX Cross Compiler Tool Chain to Path
         run: |
           echo "$OSXCROSS_DIR/target/bin" >> $GITHUB_PATH
 
       - name: Run GoReleaser
         id: run-goreleaser
-        uses: goreleaser/goreleaser-action@8f67e590f2d095516493f017008adc464e63adb1 # v4.1.0
+        uses: goreleaser/goreleaser-action@286f3b13b1b49da4ac219696163fb8c1c93e1200 # v4.1.0
```
abhisek, line 75: Can we please update the # 4.1.0 comment to the version that the SHA should resolve to?

abhisek (commented):
LGTM. Minor feedback added.

abhisek (approved):

PR #260, opened 2024-10-20
Feat/terraform support

insaaniManav:


pkg/parser/parser.go
```diff
@@ -49,13 +51,15 @@ var supportedEcosystems map[string]bool = map[string]bool{
...
 	models.EcosystemGitHubActions: true,
+	models.EcoSystemTerraform:     true,
 }
 
 // TODO: Migrate these to graph parser
 var customExperimentalParsers map[string]lockfile.PackageDetailsParser = map[string]lockfile.PackageDetailsParser{
 	customParserTypePyWheel: parsePythonWheelDist,
 	customParserSpdxSBOM:    spdx.Parse,
 	customParserTypeSetupPy: py.ParseSetuppy,
+	customParserTerraform:   terraform.ParseTerraformLockfile,
```
abhisek, line 62: I think your parser can be reliable and need not be in `experimental` list. You can move your parser to `graph` parsers which is the preferred model for implementing parser going forward. You can simply move the definition to

https://github.com/safedep/vet/blob/main/pkg/parser/parser.go#L85

pkg/parser/parser.go
```diff
@@ -240,6 +245,8 @@ func (pw *parserWrapper) Ecosystem() string {
 		return models.EcosystemMaven
 	case customParserGitHubActions:
 		return models.EcosystemGitHubActions
+	case customParserTerraform:
+		return models.EcoSystemTerraform
```
abhisek, line 249: Should be `models.EcosystemTerraform`

pkg/parser/parser.go
```diff
@@ -49,13 +51,15 @@ var supportedEcosystems map[string]bool = map[string]bool{
 	models.EcosystemCyDxSBOM:      true,
 	models.EcosystemSpdxSBOM:      true,
 	models.EcosystemGitHubActions: true,
+	models.EcoSystemTerraform:     true,
 }
 
 // TODO: Migrate these to graph parser
 var customExperimentalParsers map[string]lockfile.PackageDetailsParser = map[string]lockfile.PackageDetailsParser{
```
abhisek, line 58: Add the standard filename i.e. `.terraform.lock.hcl` to `dependencyGraphParsers` map so that this parser can automatically be discovered based on filename.

`parser.FindParser(..)` is used by other packages to see if a parser exists for a given file. `FindParser(..)` in turn looks up `dependencyGraphParsers` to find parsers for standard filenames.

pkg/parser/custom/terraform/terraform.go
```diff
@@ -0,0 +1,54 @@
...
+	"fmt"
+	"github.com/google/osv-scanner/pkg/lockfile"
+	"github.com/hashicorp/hcl/v2/hclsyntax"
+	"github.com/safedep/vet/pkg/models"
+	"os"
+
+	"github.com/hashicorp/hcl/v2/hclparse"
+)
+
+func ParseTerraformLockfile(path string) ([]lockfile.PackageDetails, error) {
```
abhisek, line 13: If you are moving this to a `graph` parser model, then the function signature should be slightly different. The `graph` parser model is more extensible i.e. we can provide config params to the parser from command line args if required in future. Unless the parser is "really" experimental and unreliable, I would suggest that we implement all parser following the `graph` parser model.

See: https://github.com/safedep/vet/blob/main/pkg/parser/npm_graph.go#L34

pkg/parser/custom/terraform/terraform.go
```diff
@@ -0,0 +1,54 @@
...
+		return nil, fmt.Errorf("failed to parse lockfile: %v", diags)
+	}
+
+	body, ok := hclFile.Body.(*hclsyntax.Body)
+	if !ok {
+		return nil, fmt.Errorf("failed to assert body as hclsyntax.Body")
+	}
+	var packageDetails []lockfile.PackageDetails
+	for _, block := range body.Blocks {
+		if block.Type == "provider" {
```
abhisek, line 34: Suggestion: To simplify readability, may be you can do

```
if block.Type != "provider" {
    continue
}
```

Personally I feel less nesting is better readability.

pkg/parser/custom/terraform/terraform.go
```diff
@@ -0,0 +1,54 @@
...
+			providerName := block.Labels[0] // The provider name is the first label
+			var providerVersion string
+			if versionAttr, exists := block.Body.Attributes["version"]; exists {
```
abhisek, line 37: What if `version` does not exist? I think we should consider a default version `0.0.0` that can be checked for `zero value` if required.

pkg/parser/parser.go
insaaniManav: The reason I moved this to experimental was because I thought that it needed to be tested maybe in prod a few times and then we will move to graph , will make this change

pkg/parser/parser.go
```diff
@@ -49,13 +51,15 @@ var supportedEcosystems map[string]bool = map[string]bool{
 	models.EcosystemCyDxSBOM:      true,
 	models.EcosystemSpdxSBOM:      true,
 	models.EcosystemGitHubActions: true,
+	models.EcoSystemTerraform:     true,
```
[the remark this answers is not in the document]
insaaniManav: Wilco

insaaniManav:
@abhisek I have pushed the PR review changes, can you please do a second round of review, thanks

pkg/parser/terraform.go
```diff
@@ -0,0 +1,46 @@
+package parser
```
abhisek, line 1: @insaaniManav Can we please have tests for this please? Once the tests are written, we are good to merge. Refer to the npm parser test case if you need.

https://github.com/safedep/vet/blob/main/pkg/parser/npm_graph_test.go

pkg/parser/parser.go
```diff
@@ -83,6 +85,7 @@ type dependencyGraphParser func(lockfilePath string, config *ParserConfig) (*mod
 
 // Maintain a map of lockfileAs to dependencyGraphParser
 var dependencyGraphParsers map[string]dependencyGraphParser = map[string]dependencyGraphParser{
+	".terraform.lock.hcl":             parseTerraformLockfile,
```
abhisek, line 88: This looks a bit odd. May be we should just use `customParserTerraform` here instead of `.terraform.lock.hcl`. This map also is used when I do something like:

```
vet scan -D /path/to/dir --type terraform
```

The `type` arg is used to lookup the "explicitly provided" parser to parse input files. However, when such explicit type is missing, then we need a way to find the parser based on file type. Lets figure that out as a future enhancement. May be you can file an issue for the future.

insaaniManav: so finally should we make a change to this ?

abhisek: @insaaniManav Set this as `customParserTerraform` for this PR.

abhisek (approved):

Issue #286, opened 2024-11-23
Add Support for SPDX License Expressions


abhisek:
`vet` should support handling license expressions such as `(MIT OR CC0-1.0)` as per SPDX license expressions rule.

License expressions
https://spdx.github.io/spdx-spec/v2.3/SPDX-license-expressions/

Example:
https://www.npmjs.com/package/type-fest

This information should be used to evaluate if a policy violates the license expression for a given OSS package

harisudarsan1:
### Problem Statement

To enable SPDX expression validation in Safedep, we need to parse and evaluate license expressions against policy-defined rules. The solution must accurately handle boolean logic, versioning, and containment checks without generating false positives or negatives.

### Solution Approaches

### Approach 1: Leveraging Existing Go Libraries (spdx-exp)

- **Library**: [[spdx-exp (Go)](https://github.com/licenseware/spdx-exp)](https://github.com/licenseware/spdx-exp)
- **Capabilities**:
    - Validates license expressions.
    - Supports versioning for accurate SPDX license matching.
    - `Satisfy` function checks if a list of licenses satisfies a license expression.
- **Limitations**:
    - **Expression Comparison**: The `Satisfy` function compares an expression against a list of licenses, not another expression. This introduces issues when boolean logic is involved (e.g., `MIT AND Apache-2.0` is simplified to `[MIT, Apache-2.0]`, losing the logical structure).
    - **False Positives**: License extraction from expressions ignores boolean operators, leading to incorrect validation.
- **Key Issue**: While version-aware, the inability to compare two full expressions directly (with boolean logic intact) makes this unsuitable for robust policy validation.

### Approach 2: Custom Parsing and Comparison Utilities

- **Reference Library**: [[license-expression (Python)](https://github.com/nexB/license-expression)](https://github.com/nexB/license-expression)
- **Advantages**:
    - Better API for parsing and comparing expressions.
    - Handles boolean logic and expression normalization.
- **Limitations**:
    - Uses substring matching during containment checks, leading to false negatives (e.g., `MIT AND Apache-2.0` != `Apache-2.0 AND MIT`).
    - Version mismatches and operand reordering can cause validation failures.
- **Key Issue**: Simplification and substring-based containment checks do not fully respect SPDX versioning and boolean logic integrity.

### Recommended Approach: Custom Utilities for SPDX Expression Comparison

- **Why Custom Utilities?**
    - Existing solutions fall short in handling boolean logic, versioning, and operand reordering.
[...]

PR #304, opened 2025-01-03
Add SPDX Expression Support for License Validation in Vet  

harisudarsan1:
### **Description:**  
This PR introduces support for SPDX expressions in Vet, enabling the validation of SPDX license expressions and providing the ability to check for license containment. This feature allows users to filter packages based on their SPDX-compliant license expressions.  This PR solves #286 

### **Key Features:**  
- **SPDX Expression Parsing:** Vet now supports SPDX license expressions for package scanning and validation.  
- **License Containment Checks:** Users can filter packages by verifying if the package's license satisfies or contains a specific license expression.  
- **Command Example:**   
```bash
vet scan --purl pkg:/npm/type-fest@4.31.0 --filter 'licenses.contains_license("MIT")'
```  
The command scans the specified package (`type-fest`) and filters results to show only those that have an MIT-compatible license.  

TODOS

- [ ] Add unit tests
- [ ] Test with multiple license expressions


abhisek:
@harisudarsan1 I think it will be good to add an example in `README.md` as well specifically about license evaluation using SPDX license expressions. Somewhere here: https://github.com/safedep/vet?tab=readme-ov-file#policy-as-code

pkg/analyzer/filter/eval.go
```diff
@@ -55,8 +60,37 @@ func NewEvaluator(name string, ignoreError bool) (Evaluator, error) {
 		cel.Variable(filterInputVarScorecard, cel.DynType),
 		cel.Variable(filterInputVarLicenses, cel.DynType),
 		cel.Variable(filterInputVarRoot, cel.DynType),
+		cel.Function("contains_license",
+			cel.MemberOverload("list_string_contains_license_string", []*cel.Type{cel.ListType(cel.StringType), cel.StringType}, cel.BoolType,
+				cel.BinaryBinding(func(lhs, rhs ref.Val) ref.Val {
+					l := lhs.(traits.Lister)
```
abhisek, line 66:
```suggestion
					l, ok := lhs.(traits.Lister)
					if !ok {
					    // Return error
					}
```

pkg/analyzer/filter/eval.go
```diff
@@ -55,8 +60,37 @@ func NewEvaluator(name string, ignoreError bool) (Evaluator, error) {
...
+					filterlicenseexp := fmt.Sprintf("%s", rhs)
+					iter := l.Iterator()
+					contains := false
+					i := 0
+					for {
+						if contains {
+							break
+						}
+						if iter.HasNext().Value() == true {
```
abhisek, line 75: Fail fast is a strong engineering practice to follow. This means:

```
if !if iter.HasNext().Value() {
   break
}
```

pkg/analyzer/filter/eval.go
```diff
@@ -55,8 +60,37 @@ func NewEvaluator(name string, ignoreError bool) (Evaluator, error) {
...
+
+							satisfied, _ := spdxexp.Satisfies(filterlicenseexp, extracted)
+							contains = satisfied
+
+							i++
+						} else {
+							break
+						}
+					}
+					return types.Bool(contains)
```
abhisek, line 89:
```suggestion

					return types.Bool(contains)
```

pkg/analyzer/filter/eval.go
```diff
@@ -55,8 +60,37 @@ func NewEvaluator(name string, ignoreError bool) (Evaluator, error) {
...
+					i := 0
+					for {
+						if contains {
+							break
+						}
+						if iter.HasNext().Value() == true {
+							str := l.Get(types.Int(i))
+
+							licenseexp := fmt.Sprintf("%s", str)
+							extracted, _ := spdxexp.ExtractLicenses(licenseexp)
```
abhisek, line 79: Do not skip error handling using `_`. Handle error by failing. I think `return fmt.Errorf` is appropriate here but you should read: https://go.dev/blog/error-handling-and-go

pkg/analyzer/filter/eval.go
```diff
@@ -55,8 +60,37 @@ func NewEvaluator(name string, ignoreError bool) (Evaluator, error) {
...
+
+							satisfied, _ := spdxexp.Satisfies(filterlicenseexp, extracted)
```
abhisek, line 81: Do not skip error handling using `_`. Fail on error

pkg/analyzer/filter/eval.go
harisudarsan1: I have handled it by fatalf from logger as that function only expects boolean return value

harisudarsan1:
Maybe we can have testing as a separate issue as it involves other functions as well what do you think @abhisek

abhisek:
> Maybe we can have testing as a separate issue as it involves other functions as well what do you think @abhisek

Nope :) Writing test cases is part of shipping a feature. Lets pair up on Monday and get this to closure together. I think most of the work is done, we just need to take it to closure. I can help.

abhisek:
@harisudarsan1 I have added test cases for the filter evaluator to ensure both current and new behaviour is supported. I have reverted the changes in `samples/` because we will first use this feature ourselves in our own repositories to ensure it works fine before making it generally recommended for our users. We have updated `README.md` to mention about this change anyway, so anyone who need it can use this feature.

abhisek (approved):

PR #305, opened 2025-01-08
Minor bugfix to handle cases where the package name doesn't resolve properly

### LINKED ISSUE #300: Lockfile Poisoning Analyser Bug in Path Handling
There is a case where `strip-ansi-cjs` resolves to `https://registry.npmjs.org/strip-ansi/-/strip-ansi-6.0.1.tgz`. This is a potential Lockfile Poisoning attempt because the package name does not match the conventional URL for the artifact. We need to do the following

1. Review `npm` specification on various conventions for URLs
2. Update `--trusted-registries` option handling to handle such conventions

Currently `vet` throws error

```
** Package `strip-ansi-cjs` resolved to an URL `https://registry.npmjs.org/strip-ansi/-/strip-ansi-6.0.1.tgz` that does 
not follow the package name path convention
** Package `wrap-ansi-cjs` resolved to an URL `https://registry.npmjs.org/wrap-ansi/-/wrap-ansi-7.0.0.tgz` that does not
follow the package name path convention
** Package `string-width-cjs` resolved to an URL `https://registry.npmjs.org/string-width/-/string-width-4.2.3.tgz` that
does not follow the package name path convention
```

insaaniManav:
Fixes #300

### DIFF
--- a/pkg/analyzer/lfp_npm.go
+++ b/pkg/analyzer/lfp_npm.go
@@ -115,7 +115,7 @@ func (npm *npmLockfilePoisoningAnalyzer) Analyze(manifest *models.PackageManifes
 
 		trustedRegistryUrls := []string{npmRegistryTrustedUrlBase}
 		trustedRegistryUrls = append(trustedRegistryUrls, npm.config.TrustedRegistryUrls...)
-
+		userTrustUrls := npm.config.TrustedRegistryUrls
 		logger.Debugf("npmLockfilePoisoningAnalyzer: Analyzing package [%s] with %d trusted registry URLs in config",
 			packageName, len(trustedRegistryUrls))
 
@@ -147,7 +147,7 @@ func (npm *npmLockfilePoisoningAnalyzer) Analyze(manifest *models.PackageManifes
 			})
 		}
 
-		if !npmIsUrlFollowsPathConvention(lockfilePackage.Resolved, packageName, trustedRegistryUrls) {
+		if !npmIsUrlFollowsPathConvention(lockfilePackage.Resolved, packageName, trustedRegistryUrls, userTrustUrls) {
 			logger.Debugf("npmLockfilePoisoningAnalyzer: Package [%s] resolved to an unconventional URL [%s]",
 				packageName, lockfilePackage.Resolved)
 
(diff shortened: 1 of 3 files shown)


pkg/analyzer/lfp_npm.go
```diff
@@ -284,6 +284,17 @@ func npmIsUrlFollowsPathConvention(sourceUrl string, pkg string, trustedUrls []s
 
 	// Example: @angular/core from https://registry.npmjs.org/@angular/core/-/core-1.0.0.tgz
 	scopedPackageName := strings.Split(path, "/-/")[0]
+	if slices.Contains(acceptablePackageNames, scopedPackageName) {
+		return true
+	} else {
```
abhisek, line 289: Can we simplify this? Instead of using `else`, just fall through to the `for` loop to check for acceptable conditions. If nothing matches, `return false` as the default fallback.

pkg/analyzer/lfp_npm.go
```diff
@@ -284,6 +284,17 @@ func npmIsUrlFollowsPathConvention(sourceUrl string, pkg string, trustedUrls []s
...
+		// Check if resolved URL starts with any trusted URL
+		// This is used to handle cases where strip-ansi-cjs resolves to https://registry.npmjs.org/strip-ansi/-/strip-ansi-6.0.1.tgz
+		for _, trustedUrl := range trustedUrls {
+			if strings.HasPrefix(sourceUrl, trustedUrl) {
```
abhisek, line 293: Don't think this is correct. For example `https://registry.npmjs.org` is a trusted URL. Within this, everything will match including `https://registry.npmjs.org/x/y/z` which may not be desirable. For example, this E2E test case caught the problem:

https://github.com/safedep/vet/blob/main/test/scenarios/fixtures/lockfiles/nestjs-lfp-package-lock.json#L68

May be we need to do this check only with `npm.config.TrustedRegistryUrls` that are explicitly supplied by the users.

abhisek:
@insaaniManav Can you rebase with `main` please. Want to ensure CI is passing with latest base.

pkg/analyzer/lfp_npm.go
```diff
@@ -265,25 +264,35 @@ func npmIsUrlFollowsPathConvention(sourceUrl string, pkg string, trustedUrls []s
...
+	// Extract the scoped package name
 	scopedPackageName := strings.Split(path, "/-/")[0]
+	if slices.Contains(acceptablePackageNames, scopedPackageName) {
+		return true
+	}
 
-	return slices.Contains(acceptablePackageNames, scopedPackageName)
+	// Check if the source URL starts with any trusted URL except the NPM trusted base URL
+	for _, trustedUrl := range trustedUrls {
+		if trustedUrl == npmRegistryTrustedUrlBase {
```
abhisek, line 288: This seems like a unusual check because we explicitly call the `npmRegistryTrustedUrlBase` as trusted yet we are not trusting it. I think we should iterate over `config.TrustedRegistryUrls` so that we don't have to handle this case and to make this explicit for configured URL bases only.

abhisek:
@insaaniManav The CI was failing due to a flaky test that involves GH API. I re-ran the CI failed tests and it seems to be green

insaaniManav:
Cool can you review the code ? This is done from my end

abhisek (approved):

PR #310, opened 2025-01-16
Added php lockfile parser

### LINKED ISSUE #257: Support PHP Composer as a Lockfile Format

insaaniManav:
Fixes #257

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -23,6 +23,7 @@ const (
 	customParserTypeJavaWebAppArchive = "war"
 	customParserGitHubActions         = "github-actions"
 	customParserTerraform             = "terraform"
+	customParserComposer              = "composer"
 )
 
 var (
@@ -51,6 +52,7 @@ var supportedEcosystems map[string]bool = map[string]bool{
 	models.EcosystemSpdxSBOM:      true,
 	models.EcosystemGitHubActions: true,
 	models.EcosystemTerraform:     true,
+	models.EcosystemPHPComposer:   true,
 }
 
 // TODO: Migrate these to graph parser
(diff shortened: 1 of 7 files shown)


pkg/parser/parser.go
```diff
@@ -259,6 +263,8 @@ func (pw *parserWrapper) Ecosystem() string {
 		return models.EcosystemGitHubActions
 	case customParserTerraform:
 		return models.EcosystemTerraform
+	case customParserComposer:
+		return models.EcosystemPHPComposer
```
abhisek, line 267: It should be packagist ecosystem

pkg/parser/parser.go
```diff
@@ -51,6 +52,7 @@ var supportedEcosystems map[string]bool = map[string]bool{
 	models.EcosystemSpdxSBOM:      true,
 	models.EcosystemGitHubActions: true,
 	models.EcosystemTerraform:     true,
+	models.EcosystemPHPComposer:   true,
```
abhisek, line 55: This is not required. We should just use the `Packagist` ecosystem

abhisek (approved):

abhisek:
@insaaniManav As I read more about Composer, I see there are two files

- `composer.json`
- `composer.lock`

This is similar to `npm` where `.json` is used to describe dependencies and `.lock` to lock them to specific version. It seems like we already support `composer.lock` due to dependency on `osv-scanner` `lockfile` package which supports it. However we do not support `composer.json` because of the complexity that it does not contain exact versions of a dependency but version constraints / ranges. In such a case we need to find out a way to _choose_ a suitable version for scanning in the version range.

I think we should spend some time reading about Composer and decide on how to handle it.

insaaniManav:
According to my research
We have 2 choices 
- Best Effort Resolution:
  - Use the latest available version that satisfies the range (could lead to false negatives).
  - Use the oldest version (could lead to more vulnerabilities being flagged).
  - Allow users to choose a resolution strategy (latest, oldest, median).
- Force user to run composer install Before Scanning:
  - This would generate a composer.lock, ensuring accurate versions.

PR #335, opened 2025-02-03
fix: Ensure version is available for sync reporter

abhisek:


arunanshub (approved):

scan.go
```diff
@@ -429,6 +429,11 @@ func internalStartScan() error {
 	}
 
 	if syncReport {
+		if version == "" {
+			return fmt.Errorf("version is required for sync report, install vet using supported method: " +
+				"https://docs.safedep.io/quickstart/")
+		}
+
```
OmkarPh, line 436: Should we use cobra's pre run checks to fail-fast during creation phase itself, for such checks ?

OmkarPh: eg.
```go
cmd.PreRunE = func(cmd *cobra.Command, args []string) error {
	if syncReport && version == "" {
		return fmt.Errorf("version is required for sync report, install vet using supported method: " +
			"https://docs.safedep.io/quickstart/")
	}
	return nil
}
```

abhisek: @OmkarPh Good point. I have moved the validation to `PreRunE` cobra hook.

PR #336, opened 2025-02-03
Integrated depsusage data prepared by code analysis and report unused deps in summary

### LINKED ISSUE #308: Integrate Dependency Usage Plugin for Evidence Analysis
**Update:** We are taking a decoupled approach for integrating code analysis with vet

## Problem

A 3rd party dependency may be referenced in a lockfile such as `requirements.txt` or `package-lock.json` but not actually used in code. Vulnerabilities or other risks in such plugins are not useful.

## Background

Our code analysis framework now supports dependency usage evidence collection for Python and can be extended for other languages. https://github.com/safedep/code

## Requirements

Integrate https://github.com/safedep/code with `vet` to support `--dependency-usage-evidence` flag to enable code analysis during `scan`. For example:

```bash
vet scan -D /app/code --dependency-usage-evidence
```

This will enable code analysis during the scan. When directory is not provided using `-D` and instead a specific lockfile is provided, then the user must explicitly specific code directories. Example:

```bash
vet scan -M /path/to/code/package-lock.json --dependency-usage-evidence --code-dir /path/to/code/1 --code-dir /path/to/code/2
```

Activate the dependency usage plugin by default when `--dependency-usage-evidence` flag is provided which basically activates code analysis in `vet`. The code analysis plugins should run in parallel with the manifest scanner available in `pkg/scanner` package.

Dependency usage plugin should add necessary usage information to a `models.Package` by extending `models/models.go` as required to hold additional metadata produced by code analysis (dependency usage).

Note: We will enhance reporting as a future enhancement to leverage the dependency usage evidences from `models.Package` which holds all OSS package metadata in `vet`

OmkarPh:
Fixes #308

Try -
```bash
go run . -d code scan --app /path/to/code --db /path/to/sqlite.db
go run . -d -l - scan -D /path/to/code --dependency-usage-evidence --code-db /path/to/sqlite.db

```


abhisek:
Looks like our E2E have become flaky. Need to investigate.

pkg/models/models.go
```diff
@@ -336,6 +337,11 @@ type MalwareAnalysisResult struct {
 	VerificationRecord *malysisv1.VerificationRecord
 }
 
+type CodeAnalysisResult struct {
+	// Usage evidences of a package obtained by depsusage plugin
+	UsageEvidences []*ent.DepsUsageEvidence `json:"usage_evidences"`
```
abhisek, line 342: Hope `ent` models are JSON serializable. Do verify:

```
vet scan ... --json-dump-dir /tmp/dump
```

Open the JSON file inside the `/tmp/dump` and verify that usage evidences are available in the JSON. The reason why we don't need a specific test case here is JSON serialization is an `ent` concern. We just need to verify once

pkg/reporter/summary.go
```diff
@@ -262,6 +266,12 @@ func (r *summaryReporter) processForMalware(pkg *models.Package) {
 	}
 }
 
+func (r *summaryReporter) processForDepsUsageEvidence(pkg *models.Package) {
+	if pkg.CodeAnalysis == nil || len(pkg.CodeAnalysis.UsageEvidences) == 0 {
```
abhisek, line 270: This is probably misleading. Consider the case where `--code` is not enabled during scan. Also there are cases of transitive dependencies. We are only looking at App (1st party) code so we cannot be certain that a transitive dependency is used or not.

I think we should introduce:

- `used`
- `unknown`

Track `used` when we have usage evidence. Else unknown.

pkg/reporter/summary.go
```diff
@@ -727,6 +737,11 @@ func (r *summaryReporter) provenanceStatement() string {
 		r.summary.provenance.none)
 }
 
+func (r *summaryReporter) depsusageEvidenceStatement() string {
+	return fmt.Sprintf("%d/%d libraries unused",
```
abhisek, line 741: We should probably state that `X` libraries has usage evidence among `Y` libraries. I think we are not yet there where we can "factually state" if a library is unused. We are looking for usage evidence so we can only "state" that as truth.

pkg/scanner/enrich_codeanalysis.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/safedep/vet/pkg/code"
+	"github.com/safedep/vet/pkg/models"
+)
+
+type CodeAnalysisEnricherConfig struct {
+	ReaderRepository        code.ReaderRepository
```
abhisek, line 12: `ReaderRepository` is not a config I think. Its more of a dependency to `NewCodeAnalysisEnricher`. I think `vet` probably have some inconsistency between config and dependency but lets get it right here.

`ReaderRepository` can be passed as an arg to `NewCodeAnalysisEnricher` which explicitly states that `ReaderRepository` is a dependency for this enricher along with config.

scan.go
```diff
@@ -38,6 +40,8 @@ var (
 	scanExclude                    []string
 	transitiveAnalysis             bool
 	transitiveDepth                int
+	dependencyUsageEvidence        bool
+	codeAnalysisDbPath             string
```
abhisek, line 44:
```suggestion
	codeAnalysisDBPath             string
```

scan.go
```diff
@@ -167,6 +173,13 @@ func newScanCommand() *cobra.Command {
 	cmd.Flags().DurationVarP(&malwareAnalysisTimeout, "malware-analysis-timeout", "", 5*time.Minute,
 		"Timeout for malicious package analysis")
 
+	// mark code-db flag required only when dependencyUsageEvidence is enabled
+	cmd.PreRun = func(cmd *cobra.Command, args []string) {
```
abhisek, line 177: This is good. But if this function returns an `error` then `cobra` shows the entire help after printing the error. This is a confusing user experience because the user will have to scroll back to find the error. I think we can do something like this:

https://github.com/safedep/vet/pull/335/files

scan.go
```diff
@@ -118,6 +122,8 @@ func newScanCommand() *cobra.Command {
 		"Analyze transitive dependencies")
 	cmd.Flags().IntVarP(&transitiveDepth, "transitive-depth", "", 2,
 		"Analyze transitive dependencies till depth")
+	cmd.Flags().BoolVarP(&dependencyUsageEvidence, "dependency-usage-evidence", "", false, "Enable dependency usage evidence during scan")
+	cmd.Flags().StringVarP(&codeAnalysisDbPath, "code-db", "", "", "Path to code analysis database generated by 'vet code scan'")
```
abhisek, line 126: I am thinking if we should just call this arg `--code`. Providing a `--code /path/to/code.db` is like explicitly providing & enabling code analysis based on `code.db` during scan.

scan.go
```diff
@@ -118,6 +122,8 @@ func newScanCommand() *cobra.Command {
 		"Analyze transitive dependencies")
 	cmd.Flags().IntVarP(&transitiveDepth, "transitive-depth", "", 2,
 		"Analyze transitive dependencies till depth")
+	cmd.Flags().BoolVarP(&dependencyUsageEvidence, "dependency-usage-evidence", "", false, "Enable dependency usage evidence during scan")
```
abhisek, line 125: Should we name code related config flags as `--code-dependency-usage-evidence`? Just to make them `intention revealing` that they are associated with code analysis?

Also I think `dependency-usage-evidence` should be `true` by default because its a fairly low cost (CPU/memory) and reliable analysis.

scan.go
```diff
@@ -167,6 +173,13 @@ func newScanCommand() *cobra.Command {
 	cmd.Flags().DurationVarP(&malwareAnalysisTimeout, "malware-analysis-timeout", "", 5*time.Minute,
 		"Timeout for malicious package analysis")
 
+	// mark code-db flag required only when dependencyUsageEvidence is enabled
+	cmd.PreRun = func(cmd *cobra.Command, args []string) {
+		if dependencyUsageEvidence {
```
abhisek, line 178: If we are making `--code` as the arg to enable code analysis during `vet scan` then I don't think we need to validate this. Too many args validation makes user experience complicated. I think we should just keep options to enable / disable larger modules (e.g. code analysis) and provide opinionated defaults.

scan.go
```diff
@@ -487,6 +500,33 @@ func internalStartScan() error {
 		enrichers = append(enrichers, enricher)
 	}
 
+	if enableCodeAnalysisEnricher() {
```
abhisek, line 503: We should first check if `code` analysis is enabled by providing a DB. Then create the storage, client etc. Then based on config flags create and append the enrichers.

abhisek:
@OmkarPh Please rebase with `main`.

pkg/models/models.go
OmkarPh: yes, verified the JSON, also it has options for customisation eg. https://entgo.io/docs/schema-fields/#struct-tags

pkg/reporter/summary.go
OmkarPh: yep, makes sense

pkg/reporter/summary.go
OmkarPh: ack 👍

pkg/scanner/enrich_codeanalysis.go
OmkarPh: got it, my bad

scan.go
OmkarPh: ack

scan.go
OmkarPh: yep this one is more readable

abhisek (approved):

Issue #338, opened 2025-02-04
Add Dependency Usage Information in CSV Report


abhisek:
PR: https://github.com/safedep/vet/pull/336 introduces code analysis with `vet`.

We need to update CSV report generated using `--report-csv` with a column named `Usage in Code` and add a single sample evidence of usage in code when available. When not available, this column can be empty.

It will help users who leverage Spreadsheet for vulnerability prioritization based on usage evidence.

PR #341, opened 2025-02-04
Added usage evidence in summary and CSV report

### LINKED ISSUE #337: Add Dependency Usage Tag in vet Summary Report
PR: https://github.com/safedep/vet/pull/336 introduces code scanner with `vet`. We have dependency usage `count` integrated with Summary Reporter. We need to extend summary reporter to add `used` tag in the summary report.

Example:

Image

This table should include a tag `used-in-code` when there is such an evidence available

OmkarPh:
Fixes #337 ,  #338

image

### DIFF
--- a/pkg/reporter/csv.go
+++ b/pkg/reporter/csv.go
@@ -3,6 +3,7 @@ package reporter
 import (
 	"encoding/csv"
 	"os"
+	"strconv"
 	"strings"
 
 	"github.com/safedep/dry/utils"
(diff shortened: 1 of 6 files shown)


abhisek (dismissed):

pkg/reporter/csv.go
```diff
@@ -145,11 +148,22 @@ func (r *csvReporter) Finish() error {
...
+				}
+			}
+
 			newRecord := record
 			newRecord.osvId = vulnId
 			newRecord.cveId = cveId
 			newRecord.vulnSummary = summary
 			newRecord.vulnSeverity = risk
+			newRecord.usageEvidenceCount = usageEvidenceCount
+			newRecord.usageEvidenceSample = usageEvidenceSample
```
abhisek, line 166: Can we have an E2E test to assert for this i.e. file path? We can make it very simple and leverage the existing fixtures. The goal is to safeguard against potential future changes breaking existing contracts like what we are offering the reporting modules

abhisek (approved):

PR #345, opened 2025-02-07
Exclude patterns option & ensure code DB path exists before preparing Code DB

OmkarPh:
If user is doing - `vet code scan --db /tmp/a/b/c/dump/vet-test.db` and the intermediate dirs like `/tmp/a`, `/tmp/b` don't exists, vet would panic. This isn't handled by sqlite.
It is fixed using `os.MkdirAll` to ensure all intermediate paths exist.

Added `--exclude` mutli-string flag to skip misleading folders like `test-data/` which may contain dummy source code 
eg. 
```bash
vet code scan --db /tmp/vet-test.db --exclude test-data --exclude .venv
```

### DIFF
--- a/cmd/code/scan.go
+++ b/cmd/code/scan.go
@@ -2,6 +2,7 @@ package code
 
 import (
 	"context"
+	"regexp"
 
 	"github.com/safedep/vet/internal/command"
 	"github.com/safedep/vet/internal/ui"
@@ -15,6 +16,7 @@ var (
 	dbPath                    string
 	appDirs                   []string
 	importDirs                []string
+	excludePatterns           []string
 	skipDependencyUsagePlugin bool
 )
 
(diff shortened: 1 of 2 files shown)


cmd/code/scan.go
```diff
@@ -31,6 +33,8 @@ func newScanCommand() *cobra.Command {
 	cmd.Flags().StringVar(&dbPath, "db", "", "Path to create the sqlite database")
 	cmd.Flags().StringArrayVar(&appDirs, "app", []string{"."}, "Directories to scan for application code files")
 	cmd.Flags().StringArrayVar(&importDirs, "import-dir", []string{}, "Directories to scan for import files")
+	cmd.Flags().StringArrayVarP(&excludePatterns, "exclude", "", []string{},
```
abhisek, line 36: Whats the rationale behind this? Do we also include a `include` pattern then? If so we need to decide on the precedence

cmd/code/scan.go
```diff
@@ -59,9 +63,15 @@ func internalStartScan() error {
...
 
+	excludePatternsRegexps := []*regexp.Regexp{}
+	for _, pattern := range excludePatterns {
+		excludePatternsRegexps = append(excludePatternsRegexps, regexp.MustCompile(pattern))
+	}
+
 	codeScanner, err := code.NewScanner(code.ScannerConfig{
 		AppDirectories:            appDirs,
 		ImportDirectories:         importDirs,
+		ExcludePatterns:           excludePatternsRegexps,
```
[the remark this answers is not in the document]
OmkarPh: scanner uses to to create `fs.NewLocalFileSystem` in code, which skips these while enumerating

cmd/code/scan.go
OmkarPh: its just to ignore irrelevant things, like we have lot of code examples with vulnerabilities in say, `test/fixtures`
when building a code database, i would just do `vet code scan --app . --db vet.db --exclude test/fixtures`
so file system never enumerates these fixtures which otherwise would create false positive deps usage

abhisek (approved):

PR #368, opened 2025-03-03
feat/vscode malysis support

abhisek:
- **chore: Update API sdk to add support for vsx scanning**
- **chore: Add inspect cmd flag to skip waiting for result**
- **feat: Add support for VS Code extension reader**


OmkarPh (approved):
LGTM
We also need to enable `vsx` flag in `code` command as backlog task

PR #370, opened 2025-03-04
Updated code analysis framework

OmkarPh:
depsusage in vet scan -
code -> correctly identified exactly 4 direct deps used
vet -> identified 34 out of 41 direct deps used
kubernetes -> identified 99 out of 110 direct deps used
(total - 1549/2147 deps used)


abhisek:
@OmkarPh `ResolvePackageHint` signature is changed. Isn't it breaking vet?

OmkarPh:
Also, analysed results of kubernetes
while it identified 99/110 direct deps
it also identified usage of transitive dependencies
eg. "cel.dev/expr" is an indirect dependency - https://github.com/kubernetes/kubernetes/blob/7c78041218a4cf0b183dc0dab71f475a5db667bf/go.mod#L127

its still used in https://github.com/kubernetes/kubernetes/blob/7c78041218a4cf0b183dc0dab71f475a5db667bf/vendor/github.com/google/cel-go/common/types/types.go#L28

abhisek:
> Also, analysed results of kubernetes while it identified 99/110 direct deps it also identified usage of transitive dependencies eg. "cel.dev/expr" is an indirect dependency - https://github.com/kubernetes/kubernetes/blob/7c78041218a4cf0b183dc0dab71f475a5db667bf/go.mod#L127
> 

What are the 11 direct dependencies that we missed? Are they used in code and our dependency usage plugin missed them?

> its still used in https://github.com/kubernetes/kubernetes/blob/7c78041218a4cf0b183dc0dab71f475a5db667bf/vendor/github.com/google/cel-go/common/types/types.go#L28
> 
> If we were to rely on usage evidences to judge a dependency as direct or transitive, it won't be reliable as in this case

I would say its not a false positive from code analysis because the lib that depends on transitive dependency was vendored i.e. available in first party code base. We can probably consider a future enhancement of adding an ignore patter for FS walker in CAF

abhisek (approved):

PR #381, opened 2025-03-10
Migrated go version to 1.24

KunalSin9h:
Features

- Go Tools

Only `oapi-codegen` has been installed using `go tool` since everyother tool need to be present in the `$PATH`, like `golangci-lint` is automatically called by `lefthook`.


.github/workflows/ci.yml
```diff
@@ -16,9 +16,9 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Set up Go
-        uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568
+        uses: actions/setup-go@v5
```
abhisek, line 19: Can we please pin the action using commit SHA like `6edd4406fa81c3da01a34fa6f6343087c207a568` please? This is a security hardening best practice for GitHub Action. 

https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions

Makefile
```diff
@@ -11,17 +11,14 @@ ent:
...
 protoc-install:
 	go install google.golang.org/protobuf/cmd/protoc-gen-go@latest
 
-dev-setup: linter-install oapi-codegen-install protoc-install
+dev-setup: linter-install protoc-install
 
 oapi-codegen:
-	oapi-codegen -package insightapi -generate types ./api/insights-v1.yml > ./gen/insightapi/insights.types.go
-	oapi-codegen -package insightapi -generate client ./api/insights-v1.yml > ./gen/insightapi/insights.client.go
+	go tool oapi-codegen -package insightapi -generate types ./api/insights-v1.yml > ./gen/insightapi/insights.types.go
```
abhisek, line 20: How do we ensure specific version for `oapi-codegen`. In our case, we specifically need `v1.10.1` because other version introduces breaking change

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/safedep/vet
 
-go 1.23.2
+go 1.24
+
+tool github.com/deepmap/oapi-codegen/cmd/oapi-codegen
```
abhisek, line 5: We need to use specific version i.e. `v1.10.1`

.github/workflows/ci.yml
KunalSin9h: ok

Makefile
KunalSin9h:
```bash
go get -tool github.com/deepmap/oapi-codegen/cmd/oapi-codegen@v1.10.1
```

when installing use specify version. but we dont see versions after `tools` keyword but they are specified in `require` keyword. 

image

go.mod
KunalSin9h: yes, updated to 1.10.1

abhisek: @KunalSin9h Can we not explicitly pin the version here. Example:

```
tool github.com/deepmap/oapi-codegen/cmd/oapi-codegen@v1.10.1
```

The problem with only `go.mod` is it will get updated when we do `go get -u`. We want to be able to update all Go modules but not `oapi-codegen`

KunalSin9h: @abhisek i tried to find and even did `go get -u` and it updated the version, the whole point of using `go get -tool` is to have same level of flexibility as packages to tools, so it will update the version. 

i the new commit, i have revert to manual `go install`. let go with that.

KunalSin9h: i am investigating this, since i have found a video where `go get -u tool` is  used to updated all the tools.

abhisek: @KunalSin9h Actually we can get rid of oapi tool because we are moving away from OpenAPI to protocol buffers.

Makefile
```diff
@@ -20,8 +20,8 @@ protoc-install:
 dev-setup: linter-install oapi-codegen-install protoc-install
 
 oapi-codegen:
-	oapi-codegen -package insightapi -generate types ./api/insights-v1.yml > ./gen/insightapi/insights.types.go
-	oapi-codegen -package insightapi -generate client ./api/insights-v1.yml > ./gen/insightapi/insights.client.go
+	go tool oapi-codegen -package insightapi -generate types ./api/insights-v1.yml > ./gen/insightapi/insights.types.go
```
abhisek, line 23: Let's just remove the oapi tool and its usage from Makefile altogether because we are completely moving to gRPC.

go.mod
KunalSin9h: ok

Makefile
KunalSin9h: done

abhisek:
@KunalSin9h Looks like `vet` found some dependency with critical vulnerability being introduced. Can you check which tool / library is introducing this? We need to avoid introducing such vulnerabilities

KunalSin9h:
@abhisek i have upgraded the `risky` packages, but still we got vet error, i guess its due to `unmaintained` policy?

https://github.com/safedep/vet/actions/runs/13804427954/job/38612512531?pr=381

abhisek (approved):

PR #387, opened 2025-03-12
fix: missing version  on go install builds

### LINKED ISSUE #385: Error: version is required.. , Missing version information on vet installed by `go install`
When installing package with `go install github.com/safedep/vet@latest`, due to how go install works, version information and commit hash are missing. 

Image

> This is expected behaviors, since we aren't able to do `-ldflags -X ...` wile `go install`

But, missing version is causing other issues like *While using Cloud with `--report-sync`*

Image

> [!NOTE]
> I have seen somewhere to set version information for `go install` apps, Its not in my head but its possible.

KunalSin9h:
Fixes #385 

This solution is based on new go 1.24 feature: https://antonz.org/go-1-24/#main-modules-version

When app is build using `go install` then we will not be able to do `-ldflags`, hence our `version` will be `""`. In that case we will return version using `runtime/debugs`'s BuildInfo. 

Its value will be:

When the current commit matches a tagged version, the value is set to v<tag>[+dirty]:
```bash
v1.2.4
v1.2.4+dirty
```
When the current commit doesn't match a tagged version, the value is set to <pseudo>[+dirty], where pseudo consists of the latest tag, current date, and commit:
```bash
v1.2.3-0.20240620130020-daa7c0413123
v1.2.3-0.20240620130020-daa7c0413123+dirty
```

### DIFF
--- a/version.go
+++ b/version.go
@@ -3,6 +3,7 @@ package main
 import (
 	"fmt"
 	"os"
+	runtimeDebug "runtime/debug"
 
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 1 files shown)


KunalSin9h:
@abhisek squash it, i have refactored same thing multiple time for best result.

KunalSin9h:
Result: 

image
image

At least we will solve the issue of `invalid version` as mention on #385

abhisek:
@KunalSin9h Can you verify that this PR does not impact build with make or goreleaser and version is available when explicitly set with ldflag?

KunalSin9h:
yes, i have already tired that, when using make quick-vet, we ill never enter the code in 

```go
if version == "" {
}
```

image

PR #388, opened 2025-03-13
Implemented Defectdojo reporter

OmkarPh:
setup -
```bash
git clone https://github.com/DefectDojo/django-DefectDojo && cd django-DefectDojo
docker compose up -d
export DEFECT_DOJO_APIV2_KEY="e40....d81"

vet -d -l - scan -D <project-dir> --malware --filter 'vulns.critical.exists(p, true) || vulns.high.exists(p, true)' \
  --defect-dojo-host-url http://localhost:8080 --report-defect-dojo --defect-dojo-product-id 2
```


scan.go
abhisek, line 1: We need the add the same reporter in `query.go` so that we can do:

```
vet query --from /tmp/source --report-defect-dojo
```

scan.go
```diff
@@ -159,6 +161,8 @@ func newScanCommand() *cobra.Command {
 		"Show only packages that are used in code (requires code analysis)")
 	cmd.Flags().StringVarP(&csvReportPath, "report-csv", "", "",
 		"Generate CSV report of filtered packages")
+	cmd.Flags().BoolVarP(&reportDefectDojo, "report-defect-dojo", "", false, "Report to DefectDojo")
+	cmd.Flags().IntVarP(&defectDojoProductID, "defect-dojo-product-id", "", -1, "DefectDojo Product ID")
```
abhisek, line 165: Lets take `DEFECT_DOJO_HOST_URL` also as an arg. No point taking only that from environment. Also we should fail fast if it is not set instead of using a default `localhost` and end up failing later silently.

pkg/reporter/defect_dojo.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/safedep/vet/pkg/analyzer"
+	"github.com/safedep/vet/pkg/common/logger"
+	"github.com/safedep/vet/pkg/models"
+	"github.com/safedep/vet/pkg/policy"
+)
+
+// DefectDojo accepts findings in SARIF report format. We'll use sarfBuilder
+// to generate the SARIF report and post it to DefectDojo.
+
+const DefaultDefectDojoHostUrl = "http://localhost:8080"
```
abhisek, line 19: Not sure if this is a good default. Mostly people will run `vet` in CI and DD will be hosted remotely. In this case I think the `DefectDojoHostUrl` is a required config and we can't really fallback to a default.

OmkarPh: yep, makes sense, will update

abhisek (approved):

PR #397, opened 2025-03-17
fix uv.lock Parser Graph Root Handling

### LINKED ISSUE #394: Fix uv.lock Parser Graph Root Handling
The `uv.lock` parser wrongly considers the application itself as a package thereby breaking the feature where we group all packages by direct dependencies.

Example:

```
vet scan -M pkg/parser/fixtures/uv.lock --insights-v2 --report-summary-group-by-direct-deps
```

Results in

Image

Here `vet-uv-ex` is not a package but the name of the application as given in `uv.lock`. We should handle the case and ensure only real dependencies are parsed as packages by the `uv.lock` parser.

Sahilb315:
Fixes #394 
Updated the parser to not include root pkg (project name) & correctly get the direct dependencies


abhisek:
@OmkarPh Can you please do a review of this?

pkg/parser/uv_graph.go
```diff
@@ -22,10 +22,11 @@ type uvLockPackage struct {
...
 type uvLockPackageSource struct {
-	Type   string `toml:"type"`
-	URL    string `toml:"url"`
-	Subdir string `toml:"subdir,omitempty"`
-	Ref    string `toml:"ref,omitempty"`
+	Type    string `toml:"type"`
+	URL     string `toml:"url"`
+	Subdir  string `toml:"subdir,omitempty"`
+	Ref     string `toml:"ref,omitempty"`
+	Virtual string `toml:"virtual,omitempty"`
```
OmkarPh, line 29: i explored some examples for `uv.lock`, 
see this lock file - https://github.com/astral-sh/uv-docker-example/blob/c16a61fb3e6ab568ac58d94b73a7d79594a5d570/uv.lock#L472

It has `editable-"."` instead of virtual, can you have a look at the official spec and see what are possible values for this
I think relying solely on virtual is not enough to identify a root package

pkg/parser/uv_graph.go
```diff
@@ -71,9 +72,9 @@ func parseUvPackageLockAsGraph(lockfilePath string, config *ParserConfig) (*mode
 
 	defer func() {
 		for _, pkg := range parsedLockFile.Packages {
-			if len(pkg.Metadata.RequiresDist) != 0 {
-				for _, dep := range pkg.Metadata.RequiresDist {
-					node := uvGraphFindByVersionRange(dependencyGraph, dep.Name, dep.Specifier)
```
OmkarPh, line 76: i think `uvGraphFindByVersionRange` is no longer needed ? please remove definition fot this too

OmkarPh (changes requested):

pkg/parser/uv_graph.go
Sahilb315: Will look into more uv.lock files containing a large number of pkgs

Sahilb315:
@OmkarPh Can you review the updated changes

pkg/parser/uv_graph.go
```diff
@@ -71,9 +73,9 @@ func parseUvPackageLockAsGraph(lockfilePath string, config *ParserConfig) (*mode
...
+			if pkg.Source.Virtual != "" || pkg.Source.Editable != "" {
+				for _, dep := range pkg.Dependencies {
+					node := uvFindPackageByName(dependencyGraph, dep.Name)
 					if node != nil {
 						node.SetRoot(true)
 					}
```
OmkarPh, line 81: here, we're setting dependencies of the root package as root of our graph, shouldn't we also mark its dev dependencies as root (if `IncludeDevDependencies` is set) ?

CC @abhisek

OmkarPh: other than this, it LGTM

Sahilb315: In npm parsing also only direct dependencies are marked as root & not dev dependencies
https://github.com/safedep/vet/blob/main/pkg%2Fparser%2Fnpm_graph.go#L146

OmkarPh (approved):

pkg/parser/uv_graph.go
abhisek: LGTM

abhisek (approved):

Issue #273, opened 2024-11-11
Show Progress in Cloud Sync Reporter


abhisek:
User feedback

> Also when I ran vet with cloud sync command, it doesn't show whether the data has been uploaded to the cloud or not it shows the standered response like VET cli. Should we think about adding some additional message with successful sync when the vet cloud command is used?

We need to update cloud report syncing to have a callback with status and metrics. This callback can be used by the main scan controller (`scan.go`) to render a progress bar for cloud reporting syncing.

PR #400, opened 2025-03-18
feat: progress bar in cloud report syncing

KunalSin9h:
Fixes: #273 

Demo: 

rec

### DIFF
--- a/scan.go
+++ b/scan.go
@@ -523,6 +523,9 @@ func internalStartScan() error {
 			EnableMultiProjectSync: syncEnableMultiProject,
 			ClientConnection:       clientConn,
 		}, reporter.SyncReporterCallbacks{
+			OnSyncStart: func() {
+				ui.PrintMsg("🌐 Syncing data to SafeDep Cloud...")
+			},
 			OnPackageSync: func(pkg *models.Package) {
 				ui.IncrementTrackerTotal(syncReportTracker, 1)
 			},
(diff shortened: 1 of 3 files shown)


scan.go
```diff
@@ -668,6 +690,7 @@ func internalStartScan() error {
 		BeforeFinish: func() {
 			ui.MarkTrackerAsDone(packageManifestTracker)
 			ui.MarkTrackerAsDone(packageTracker)
+			ui.MarkTrackerAsDone(syncReportTracker)
```
abhisek, line 693: Possible `nil` deref here because we are creating `syncReportTracker` conditionally at 

https://github.com/safedep/vet/pull/400/files#diff-37272e63b3a935ee8196994289044044d85351186f29286b458a60fd513183a4R677

scan.go
```diff
@@ -668,6 +690,9 @@ func internalStartScan() error {
 		BeforeFinish: func() {
```
abhisek, line 690: There is a race condition here. This callback is called before the Scan Finish phase. 

https://github.com/safedep/vet/blob/main/pkg/scanner/scanner.go#L87

When this callback is called, we end up marking the trackers as done. But the sync reporter is still running even after that. You will experience this is a lag / delay when you are scanning a repo with a lot of packages.

This is because the scanner triggers `Finish` for all analysers and reports after that. Sync reporter in turn waits for all sync to be completed before returning from `Finish`

Look at `OnStop` callback from scanner and see if we should stop the trackers `OnStop` instead

https://github.com/safedep/vet/blob/main/pkg/scanner/scanner.go#L93

KunalSin9h: @abhisek 

Stopping triggers at `OnStop` is causing issues in UI, see the demo gif. Analysis output is print before the progress bar is stopped. 

result

KunalSin9h:
@abhisek this is the best result i am achieve, since ui is cluttering and sames thing (the progress) bar is reprinting on bottom due to handing `ui.StopProgressTrakcer` in `OnStop`. 

image

abhisek (dismissed):

PR #408, opened 2025-03-21
fix: missing shell for using vet's image as environment

KunalSin9h:
`debian:bullseye-slim` is very small, 80 MB, and current docker image is also near 80 MB.

This is critically needed as, Gitlab need image to `run as shell` for executing commands, like `vet scan` but current docker image being distroless does not have shell (sh). 

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ ENV CGO_ENABLED=1
 
 RUN make quick-vet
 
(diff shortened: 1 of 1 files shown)


arunanshub (approved):

KunalSin9h:
failing ci: https://gitlab.com/safedep/ci-components/vet/-/jobs/9480823078

PR #419, opened 2025-03-24
Feat/gitlab report

### LINKED ISSUE #404: `--report-sca-gitlab`, a new report schema used for CI Component in gitlab
Documentation: https://docs.gitlab.com/development/integrations/secure/#report

> since `vet` is SCA, we only need to see `dependency_scanning` format. 

Json Schema: https://gitlab.com/gitlab-org/security-products/security-report-schemas/-/blob/master/dist/dependency-scanning-report-format.json

Gitlab component implementation approach:  https://github.com/safedep/vet/issues/314#issuecomment-2733089009

KunalSin9h:
fixes #404 

need and results are in this notion docs: 

https://www.notion.so/safedep-inc/Need-for-GitLab-specific-schema-reporting-1c061d70b23680319849c32d2b0cbcd6?pvs=4


abhisek:
@arunanshub Can you do a first level code review of this please? Keep it open after review. I want to review as well

pkg/reporter/gitlab.go
```diff
@@ -0,0 +1,290 @@
...
+
+// GitLabReport is the struct for the GitLab, currently using the 15.2.1 schema
+// and `dependency_scanning` type.
+// but can be extended to support other types and schemas in the future.
+// docs: https://docs.gitlab.com/development/integrations/secure/#report
+type GitLabReport struct {
+	Schema  string `json:"schema"`
+	Version string `json:"version"`
+	Scan    struct {
+		Scanner struct {
```
abhisek, line 63: Can we avoid using nested structs please? Very hard to use them. Just define `Scanner` as a separate struct.

If you have generated this Go code from JSON example using the below utility then avoid using `inline type definition`.

https://mholt.github.io/json-to-go/

KunalSin9h: got it

KunalSin9h:
i have verified it working on actual ci, 
image

abhisek:
@KunalSin9h Is there a way to include Malicious Package scanning results as well in the Gitlab JSON when available in vet? How to enable that with credentials (API Key) in Gitlab is outside the scope of this PR.

When malicious package scanning is enabled, this information is available at the package level. See this code:
https://github.com/safedep/vet/blob/main/pkg/models/models.go#L369

KunalSin9h:
@abhisek 
> How to enable that with credentials (API Key) in Gitlab is outside the scope of this PR.

does this is enabled by using `--malware` and `--malware-analysis-timeout` flags? 
if so, then they are already part of gitlab ci scirpt

image

KunalSin9h:
Yes, though we dont have any extra filed in schema, these reports need to be in same `vulnerability`. 

Should i add this functionality, @abhisek 
Waht is the `URL` for any malware report, i need that.

abhisek:
> Yes, though we dont have any extra filed in schema, these reports need to be in same `vulnerability`.
> 
> Should i add this functionality, @abhisek What is the `URL` for any malware report, i need that.

@KunalSin9h Yes. Please go ahead and add malicious package scan results as vulnerabilities when available.

Malware=True CRITICAL
Suspicious=True HIGH

Here is an utility to get URL from analysisId:
https://github.com/safedep/vet/blob/main/pkg/malysis/url.go

The analysisId is available in the model
https://github.com/safedep/vet/blob/main/pkg/models/models.go#L327

KunalSin9h:
image
image

pkg/reporter/gitlab_test.go
```diff
@@ -0,0 +1,278 @@
...
+	t.Run("NewGitLabReporter", func(t *testing.T) {
+		reporter, err := NewGitLabReporter(GitLabReporterConfig{
+			Path: reportPath,
+		})
+		assert.NoError(t, err)
+		assert.NotNil(t, reporter)
+		assert.Equal(t, "GitLab Dependency Scanning Report Generator", reporter.Name())
+	})
+
+	t.Run("Empty Report", func(t *testing.T) {
```
arunanshub, line 36: Instead of writing t.Run multiple times, create a table and write a table driven test instead.

pkg/reporter/gitlab.go
```diff
@@ -0,0 +1,329 @@
...
+}
+
+type gitLabReporter struct {
+	config          GitLabReporterConfig
+	vulnerabilities []GitLabVulnerability
+	startTime       time.Time
+}
+
+// Ensure gitLabReporter implements Reporter interface
+var _ Reporter = (*gitLabReporter)(nil)
```
arunanshub, line 109: Why are we doing this?

pkg/reporter/gitlab.go
```diff
@@ -0,0 +1,329 @@
...
+				switch risk {
+				case insightapi.PackageVulnerabilitySeveritiesRiskCRITICAL:
+					severity = "Critical"
+				case insightapi.PackageVulnerabilitySeveritiesRiskHIGH:
+					severity = "High"
+				case insightapi.PackageVulnerabilitySeveritiesRiskMEDIUM:
+					severity = "Medium"
+				case insightapi.PackageVulnerabilitySeveritiesRiskLOW:
+					severity = "Low"
+				}
```
arunanshub, line 266: This can be moved to its own function

pkg/reporter/gitlab.go
```diff
@@ -0,0 +1,329 @@
...
+				Version: pkg.GetVersion(),
+				Direct:  pkg.Depth == 0,
+			},
+		}
+
+		// Add malware analysis result
+		malwareAnalysis := pkg.MalwareAnalysis
+
+		if malwareAnalysis != nil && (malwareAnalysis.IsMalware || malwareAnalysis.IsSuspicious) {
+			severity := "Critical"
```
arunanshub, line 212: Instead of hardcoding this value, can we use a function to get the value? (See comment below)

pkg/reporter/gitlab.go
abhisek: @arunanshub This is an idiomatic Go way of ensuring `gitLabReporter` follows the contract defined by interface `Reporter`. While `NewGitlabReporter` can directly return `Reporter` interface, it is generally preferred to return `* gitLabReporter` to help in testing internal functions

pkg/reporter/gitlab.go
```diff
@@ -0,0 +1,346 @@
...
+			reportUrl := ""
+
+			if malwareAnalysis.Report != nil {
+				reportUrl = malysis.ReportURL(malwareAnalysis.Report.ReportId)
+				if malwareAnalysis.Report.Inference != nil {
+					description = fmt.Sprintf("%s\n\n%s", malwareAnalysis.Report.Inference.Summary, malwareAnalysis.Report.Inference.Details)
+				}
+			}
+
+			malwareId := fmt.Sprintf("MAL-%s", malwareAnalysis.AnalysisId)
```
abhisek, line 238: We are generating an opinionated ID here. To maintain consistency anywhere else we use this ID, we should define this at the model itself so that you can call `.Id()` here. Also add a comment where you define `Id()` that its opinionated and is not linked to any standard IDs

pkg/reporter/gitlab.go
```diff
@@ -0,0 +1,346 @@
...
+			glVuln := GitLabVulnerability{
+				ID:          malwareId,
+				Name:        fmt.Sprintf("%s@%s is Malware/Suspicious Package", pkg.GetName(), pkg.GetVersion()),
+				Description: description,
+				Severity:    severity,
+				Location:    location,
+				Identifiers: []GitLabIdentifier{
+					{
+						Type:  "malware",
```
abhisek, line 247: `"malware"` should be defined as a constant if it has meaning in Gitlab JSON schema

pkg/reporter/gitlab.go
```diff
@@ -0,0 +1,346 @@
...
+
+			summary := utils.SafelyGetValue(vulns[i].Summary)
+			// Create GitLab vulnerability entry
+			glVuln := GitLabVulnerability{
+				ID:          utils.SafelyGetValue(vulns[i].Id),
+				Name:        summary,
+				Description: summary, // Using summary as description since that's what we have
+				Severity:    severity,
+				Location:    location,
+				// Todo: Solution
```
abhisek, line 279: Lets remove unnecessary comments and create issues in our backlog. Comments are hard to track and action upon in future.

pkg/reporter/gitlab.go
```diff
@@ -0,0 +1,346 @@
...
+}
+
+func (r *gitLabReporter) AddAnalyzerEvent(event *analyzer.AnalyzerEvent) {}
+
+func (r *gitLabReporter) AddPolicyEvent(event *policy.PolicyEvent) {}
+
+func (r *gitLabReporter) Finish() error {
+	vendor := GitLabVendor{Name: "safedep"}
+	scanner := GitLabScanner{
+		ID:      "vet",
```
abhisek, line 298: Should be defined as constant or include as part of config similar to what we have done here:
https://github.com/safedep/vet/blob/main/pkg/reporter/sync.go#L53

pkg/reporter/gitlab.go
```diff
@@ -0,0 +1,346 @@
...
+		Name:    "vet",
+		Version: r.config.VetVersion,
+		Vendor:  vendor,
+	}
+
+	report := GitLabReport{
+		Schema:  "https://gitlab.com/gitlab-org/security-products/security-report-schemas/-/raw/15.2.1/dist/dependency-scanning-report-format.json",
```
abhisek, line 305: Should be a constant

pkg/reporter/gitlab.go
```diff
@@ -0,0 +1,346 @@
...
+		Version: "15.2.1",
```
abhisek, line 306: What is the significance of this version? How does it change? This should be a constant at the very least unless there is a "logic" that leads to this version

pkg/reporter/gitlab.go
```diff
@@ -0,0 +1,346 @@
...
+		Scan: GitLabScan{
+			Scanner:   scanner,
+			Analyzer:  scanner, // Using same scanner info for analyzer
+			Type:      "dependency_scanning",
+			StartTime: formatTime(r.startTime),
+			EndTime:   formatTime(time.Now()),
+			Status:    "success",
```
abhisek, line 313: Is `"success"` meaningful in Gitlab schema? If so, move this as a constant

pkg/reporter/gitlab.go
```diff
@@ -0,0 +1,346 @@
...
+		Vendor:  vendor,
+	}
+
+	report := GitLabReport{
+		Schema:  "https://gitlab.com/gitlab-org/security-products/security-report-schemas/-/raw/15.2.1/dist/dependency-scanning-report-format.json",
+		Version: "15.2.1",
+		Scan: GitLabScan{
+			Scanner:   scanner,
+			Analyzer:  scanner, // Using same scanner info for analyzer
+			Type:      "dependency_scanning",
```
abhisek, line 310: Define a constant for this

pkg/reporter/gitlab.go
```diff
@@ -0,0 +1,346 @@
...
+	}, nil
+}
+
+func (r *gitLabReporter) Name() string {
+	return "GitLab Dependency Scanning Report Generator"
+}
+
+// GitLab requires time to be in this format
+// Learned the hard way :), (not that actually , thanks to Cursor)
+func formatTime(t time.Time) string {
```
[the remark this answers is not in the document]
KunalSin9h: yes, added schema reference in comments, and changed function name.

pkg/reporter/gitlab.go
```diff
@@ -0,0 +1,346 @@
...
+		}
+	}
+
+	// Add identifiers in order of priority
+	identifiers := make([]GitLabIdentifier, 0)
+
+	// Add all CVEs first
+	for _, cve := range cves {
+		identifiers = append(identifiers, GitLabIdentifier{
+			Type:  "cve",
```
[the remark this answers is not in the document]
KunalSin9h: yes, they are constant have predefined types. I have only included 3 identifiers, but now i will do for all .


image

pkg/reporter/gitlab.go
```diff
@@ -0,0 +1,346 @@
...
+	"github.com/safedep/vet/pkg/policy"
+)
+
+type GitLabReporterConfig struct {
+	Path       string // Report path, value of --report-gitlab
+	VetVersion string // Vet version, value from version.go
+}
+
+// GitLabVendor represents vendor information
+type GitLabVendor struct {
```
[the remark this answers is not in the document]
KunalSin9h: done

pkg/reporter/gitlab.go
```diff
@@ -0,0 +1,346 @@
...
+
+		// Package location
+		location := GitLabLocation{
+			File: manifest.Path,
+			Dependency: GitLabDependency{
+				Package: GitLabPackage{
+					Name: pkg.GetName(),
+				},
+				Version: pkg.GetVersion(),
+				Direct:  pkg.Depth == 0,
```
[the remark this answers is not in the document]
KunalSin9h: done

pkg/reporter/gitlab.go
```diff
@@ -0,0 +1,346 @@
...
+				Location:    location,
+				Identifiers: []GitLabIdentifier{
+					{
+						Type:  "malware",
+						Name:  malwareId,
+						Value: malwareId,
+						URL:   reportUrl,
+					},
+				},
+				// TODO
```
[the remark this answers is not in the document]
KunalSin9h: done

pkg/reporter/gitlab.go
KunalSin9h: no, gitlab has to `malware` type. its a costume type.

pkg/reporter/gitlab.go
KunalSin9h: done

pkg/reporter/gitlab.go
KunalSin9h: i have defined it as custom identifier.

pkg/reporter/gitlab.go
KunalSin9h: included as part of config

pkg/reporter/gitlab.go
KunalSin9h: Version represents the report schema version. GitLab verify syntax of our report against particular version. So if we use `schema` version `15.2.1` then we should specify exact version. 
So you can say its hard coupled with our `go struct` (the schema)

KunalSin9h: it changes when ever gitlab team releases new schema version. (to support new gitlab ui features)

pkg/reporter/gitlab.go
KunalSin9h: done

pkg/reporter/gitlab.go
KunalSin9h: done

pkg/reporter/gitlab.go
```diff
@@ -0,0 +1,346 @@
...
+		if malwareAnalysis != nil && (malwareAnalysis.IsMalware || malwareAnalysis.IsSuspicious) {
+			severity := SeverityCritical
+			if malwareAnalysis.IsSuspicious {
+				severity = SeverityHigh
+			}
+
+			description := ""
+			reportUrl := ""
+
+			if malwareAnalysis.Report != nil {
```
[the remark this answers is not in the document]
KunalSin9h: this is not method like GetReport()?

pkg/common/identifier_url.go
```diff
@@ -0,0 +1,20 @@
+package common
+
+import (
+	"fmt"
+	"strings"
+)
+
+// GetIdentifierURL returns the URL for a given identifier and value
+// Some Values need to be trimmed to use it in the URL
+func GetIdentifierURL(identifier, value string) string {
```
abhisek, line 10: This is "overloading" the purpose of this function. You should create separate functions for each. Example `GetCveReferenceURL()`, `GetCweReferenceURL` etc.

This suggestion is based on:
https://en.wikipedia.org/wiki/Single-responsibility_principle

pkg/reporter/gitlab.go
```diff
@@ -0,0 +1,358 @@
...
+type gitLabReporter struct {
+	config          GitLabReporterConfig
+	vulnerabilities []gitLabVulnerability
+	startTime       time.Time
+}
+
+// Ensure gitLabReporter implements Reporter interface
+var _ Reporter = (*gitLabReporter)(nil)
+
+func NewGitLabReporter(config GitLabReporterConfig) (Reporter, error) {
```
abhisek, line 147:
```suggestion
func NewGitLabReporter(config GitLabReporterConfig) (*gitLabReporter, error) {
```

We can return the internal type here. Since we are using `var _ Reporter = (*gitLabReporter)(nil)` to verify contract. Exposing the internal type helps in testing private functions with tests available in the same package.

pkg/reporter/gitlab.go
```diff
@@ -0,0 +1,358 @@
...
+				severity = SeverityHigh
+			}
+
+			description := ""
+			reportUrl := ""
+
+			if malwareAnalysis.Report != nil {
+				reportUrl = malysis.ReportURL(malwareAnalysis.Report.ReportId)
+				if malwareAnalysis.Report.Inference != nil {
+					description = fmt.Sprintf("%s\n\n%s", malwareAnalysis.Report.Inference.Summary, malwareAnalysis.Report.Inference.Details)
```
abhisek, line 251: Here we can use `malwareAnalysis.Report.GetInference().GetSummary()` to be nil safe. Same for `Inference.Details`

pkg/reporter/gitlab.go
```diff
@@ -0,0 +1,358 @@
...
+	// Write to file
+	err = os.WriteFile(r.config.Path, jsonData, os.ModePerm)
+	if err != nil {
+		return fmt.Errorf("failed to write GitLab report: %w", err)
+	}
+
+	return nil
+}
+
+func getVulnerabilitySeverity(risk insightapi.PackageVulnerabilitySeveritiesRisk) Severity {
```
abhisek, line 345: Can we scope it within `gitLabReporter` i.e. define it as `func (*gitLabReporter) ...` I am a bit worried about this exposed functions outside the gitlab reporter even if only within the `reporter` package

pkg/reporter/gitlab.go
```diff
@@ -0,0 +1,358 @@
...
+func (r *gitLabReporter) Name() string {
+	return "GitLab Dependency Scanning Report Generator"
+}
+
+// GitLab requires time to be in pattern
+// "^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}$"
+// Example: "2020-01-28T03:26:02"
+//
+// Docs (Schema Reference): https://gitlab.com/gitlab-org/security-products/security-report-schemas/-/blob/master/dist/sast-report-format.json#L497
+func gitlabFormatTime(t time.Time) string {
```
abhisek, line 164: If this function exists ONLY for the gitlab reporter then we should scope it within the `*gitLabReporter` struct I think?

pkg/reporter/gitlab.go
```diff
@@ -0,0 +1,358 @@
...
+	return t.Format("2006-01-02T15:04:05")
+}
+
+// gitlabAddVulnerabilityIdentifiers adds all relevant identifiers for a vulnerability
+// following GitLab's identifier guidelines
+// Docs: https://docs.gitlab.com/development/integrations/secure/#identifiers
+func gitlabAddVulnerabilityIdentifiers(vuln *gitLabVulnerability, vulnData *insightapi.PackageVulnerability) {
```
abhisek, line 171: If this function exists ONLY for the gitlab reporter then we should scope it within the `*gitLabReporter` struct I think?

pkg/reporter/gitlab.go
KunalSin9h: done

abhisek (approved):

PR #424, opened 2025-03-25
feat: test container working: scan with cloud

KunalSin9h:
locally tested ✅

approach, run docker container with cloud, store result in `scan_output.txt` file
check if the file contains the "Summary of Findings" text.


.github/workflows/ci.yml
```diff
@@ -86,3 +86,22 @@ jobs:
         docker buildx build --platform linux/amd64 --platform linux/arm64 \
         -t build-container-test .
 
+    - name: Test Container Image
+      run: |
+        docker run --rm \
```
abhisek, line 91: You need to mount `pwd` inside the container as a volume

.github/workflows/ci.yml
```diff
@@ -86,3 +86,22 @@ jobs:
...
+          build-container-test \
+          scan \
+          --filter-suite .github/vet/policy.yml \
+          --cloud-mode true \
```
abhisek, line 95: These are `vet-action` params. Not really vet. You just need to set the following env vars and then run `auth verify`. Example:

https://github.com/safedep/vet/pull/421/files

KunalSin9h:
finally

.github/workflows/ci.yml
```diff
@@ -83,6 +83,14 @@ jobs:
...
       run: |
-        docker buildx build --platform linux/amd64 --platform linux/arm64 \
-        -t build-container-test .
+        docker buildx build --platform linux/amd64 --load \
+        -t build-container-test:latest .
+
+    - name: Test Container Image
+      run: |
+        docker run --rm \
+          -e VET_API_KEY=${{ secrets.SAFEDEP_CLOUD_API_KEY }} \
```
abhisek, line 92: @KunalSin9h `secrets` are not available when a PR is raised from forked repository due to security reason. So this job will always fail when someone from the community raises a PR from forked repo.

What I suggest is you run the `Test Container Image` step only if the PR is from the same repository and not forked repository. See example:

https://github.com/safedep/vet/blob/main/.github/workflows/vet-ci.yml#L33

abhisek (approved):

PR #434, opened 2025-03-28
Implemented CycloneDX reporter with metadata, packages & vulnerabilities

OmkarPh:
Use flags -
```bash
--report-cdx path/to/cdxreport.json \
--report-cdx-app-name myapp # optional
```


pkg/common/utils/pointers.go
```diff
@@ -0,0 +1,5 @@
+package utils
+
+func PtrTo[T any](v T) *T {
```
abhisek, line 3: We have a `StringPtr` util here:

https://github.com/safedep/vet/blob/main/pkg/common/utils/string.go#L4

`PtrTo` seems more generic. Can we get rid of `StringPtr` to and use this instead? Remove `utils/string.go` if not required.

pkg/readers/github_org_reader.go
```diff
@@ -43,8 +44,13 @@ func (p *githubOrgReader) Name() string {
 	return "Github Organization Package Manifest Reader"
 }
 
+func (p *githubOrgReader) ApplicationName() (string, error) {
+	return githubOrgFromURL(p.config.OrganizationURL)
```
abhisek, line 48: This seems like misleading. Org should not be application name.Not sure what is a good default app name for a GitHub org

pkg/readers/lockfile_reader.go
```diff
@@ -26,11 +26,16 @@ func (p *lockfileReader) Name() string {
 	return "Lockfiles Based Package Manifest Reader"
 }
 
+func (p *lockfileReader) ApplicationName() (string, error) {
+	return "vet-scanned-project", nil
```
abhisek, line 30: Can we have a `const` defined for a default app name that can be used by all readers that need it? May be just have a `common.go` in `readers` package for common stuff

pkg/reporter/cyclonedx.go
```diff
@@ -0,0 +1,240 @@
...
+			Name:       config.ApplicationComponentName,
+			Components: &[]cdx.Component{},
+		},
+		Tools: &cdx.ToolsChoice{
+			Components: &[]cdx.Component{
+				{
+					Type: cdx.ComponentTypeApplication,
+					Manufacturer: &cdx.OrganizationalEntity{
+						Name: "SafeDep",
+						URL:  &[]string{"https://safedep.io/"},
```
abhisek, line 85: Use `PtrTo`?

pkg/reporter/cyclonedx.go
```diff
@@ -0,0 +1,240 @@
...
+		bom.SerialNumber = config.SerialNumber
+	}
+
+	bom.Metadata = &cdx.Metadata{
+		// Define metadata about the main component (the root component which BOM describes)
+		Component: &cdx.Component{
+			BOMRef:     "root-application",
+			Type:       cdx.ComponentTypeApplication,
+			Name:       config.ApplicationComponentName,
+			Components: &[]cdx.Component{},
```
abhisek, line 77: Use `PtrTo` util?

pkg/reporter/cyclonedx.go
```diff
@@ -0,0 +1,240 @@
...
+	SerialNumber string
+}
+
+type cycloneDXReporter struct {
+	sync.Mutex
+	config CycloneDXReporterConfig
+	bom    *cdx.BOM
+}
+
+var cdxUUIDRegexp = regex.MustCompileAndCache(`^urn:uuid:[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$`)
```
abhisek, line 49: Does the CDX spec says that the UUID need to be in this format?

pkg/common/utils/pointers.go
OmkarPh: cool, will replace with `PtrTo`

pkg/readers/github_org_reader.go
OmkarPh: hey, if we're ok with `vet-scanned-project` as default name, then maybe this can be `vet-scanned-safedep-projects` (safedep is org name)

pkg/readers/lockfile_reader.go
OmkarPh: yep, thats better

pkg/reporter/cyclonedx.go
OmkarPh: ack

pkg/reporter/cyclonedx.go
OmkarPh: ack

pkg/reporter/cyclonedx.go
OmkarPh: yes, i've taken regex from their spec - https://cyclonedx.org/docs/1.6/json/#serialNumber

pkg/readers/github_org_reader.go
abhisek: @OmkarPh Sounds good

pkg/reporter/cyclonedx.go
```diff
@@ -0,0 +1,302 @@
...
+func (r *cycloneDXReporter) AddPolicyEvent(event *policy.PolicyEvent) {
+	// @TODO - How to handle policy events
+}
+
+func (r *cycloneDXReporter) finaliseBom() {
+	bomGenerationTime := time.Now().UTC()
+
+	r.bom.Metadata.Timestamp = bomGenerationTime.Format(time.RFC3339)
+
+	annotation := (*r.bom.Annotations)[0]
```
abhisek, line 275: This seems weird. Especially since we are overwriting `r.bom.Annotations` in line number 278 anyway

OmkarPh: creating annotations during finalise phase to prevent this

abhisek:
@OmkarPh Can you rebase with `main` please?

pkg/reporter/cyclonedx.go
```diff
@@ -0,0 +1,383 @@
+package reporter
```
abhisek, line 1: @OmkarPh We should have test cases for this reporter.

abhisek (approved):

KunalSin9h:
@OmkarPh @abhisek Nice

PR #435, opened 2025-03-29
Fixed plaintext in sarif messages

OmkarPh:
Exported sarif message -
```
message": {
  "text": "Policy Violation Package langchain violates policy filter-query.",
  "markdown": "## Policy Violation\nPackage `langchain` violates policy `filter-query`.\n\n"
},
```

Reported in defectdojo -
image


pkg/reporter/markdown/markdown_test.go
```diff
@@ -63,3 +63,26 @@ func TestMarkdownBuilder(t *testing.T) {
 		})
 	}
 }
+
+func TestMarkdownBuildPlainText(t *testing.T) {
```
abhisek, line 67: Is `AAAA # BBB` valid markdown? Hope the regex won't end up replacing `#` which in this case is intended as a character within the text.

OmkarPh: yes, it won't replace if its in between, i'll add this in testcase

abhisek (approved):

PR #439, opened 2025-03-31
feat: Improve Description and Solution for Vulnerabilities

### LINKED ISSUE #428: Feat: Solution for Vul / Malware in GitLab.
In GitLab vulnerability schema, we have a text property called `solution`, this is used in the report in gitlab UI to describe solution. 

**Demo**: 

Image

**Reference**: 

Schema: https://gitlab.com/gitlab-org/security-products/security-report-schemas/-/blob/master/dist/dependency-scanning-report-format.json#L853

**Code Changes:** 

For this issue when we have solutions, we need to update `pkg/reporter/gitlab.go` 

```diff
type gitLabVulnerability struct {
	ID          string             `json:"id"`
	Name        string             `json:"name"`
	Description string             `json:"description"`
	Severity    Severity           `json:"severity"`
+       Solution    string             `json:"solution"`
	Location    gitLabLocation     `json:"location"`
	Identifiers []gitLabIdentifier `json:"identifiers"`
}
```

KunalSin9h:
Fixes #428 

Result: 

image

### DIFF
--- a/pkg/reporter/gitlab.go
+++ b/pkg/reporter/gitlab.go
@@ -385,8 +385,12 @@ func (r *gitLabReporter) getGitLabVulnerabilityDescription(pkg *models.Package, 
 // getVulnerabilitySolution returns the solution for a vulnerability
 // Markdown formatted
 func (r *gitLabReporter) getGitLabVulnerabilitySolution(pkg *models.Package) string {
-	latestVersion := utils.SafelyGetValue(pkg.Insights.PackageCurrentVersion)
-	solution := fmt.Sprintf("Upgrade to latest version **`%s`**", latestVersion)
+	solution := "No solution available for this vulnerability"
+
+	if pkg.Insights != nil && pkg.Insights.PackageCurrentVersion != nil {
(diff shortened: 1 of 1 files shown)


pkg/reporter/gitlab.go
```diff
@@ -364,3 +367,26 @@ func (r *gitLabReporter) getVulnerabilitySeverity(risk insightapi.PackageVulnera
...
+		pkgVersion,
+		summary)
+
+	return description
+}
+
+// getVulnerabilitySolution returns the solution for a vulnerability
+// Markdown formatted
+func (r *gitLabReporter) getGitLabVulnerabilitySolution(pkg *models.Package) string {
+	latestVersion := utils.SafelyGetValue(pkg.Insights.PackageCurrentVersion)
```
abhisek, line 388: There are cases where `PackageCurrentVersion` may be empty. I think we should check that and add an appropriate message saying `No solution available for this issue` to avoid incorrect advice.

abhisek (approved):

Issue #430, opened 2025-03-26
Add Vulnerabilities & Malicious Package Info in SARIF Report


abhisek:
- Add vulnerabilities to SARIF report
- Add malicious package info (if available) as critical findings in SARIF report

Note: Currently SARIF report builder only reports policy violations. To ensure the behaviour is backward compatible, introduce a configuration option for SARIF builder to include vulnerabilities and malicious package info the report.

KunalSin9h:
@abhisek current `--report-gitlab` only reports `Non-policy-violations`. (just reverse of this issue).

abhisek:
@KunalSin9h Thats fine. We will add policy violation info as a future enhancement when required.

OmkarPh:
Hey, I'll add flags `--report-sarif-malware` & `--report-sarif-vuln` to configure the sarif Reporter
Since, we're using sarif reports defectdojo also, should i enable these by default for that ?

abhisek:
@OmkarPh We can include by default i think since this is an enhancement and not a breaking change.

PR #441, opened 2025-03-31
Added vulnerabilities & malware in SARIF reports

OmkarPh:
Use flags -
```bash
--report-sarif $SAMPLE_OUTNAME.sarif.json --report-sarif-vulns --report-sarif-malware
```

Sample -
[vet-rakshak.sarif.json](https://github.com/user-attachments/files/19561722/vet-rakshak.sarif.json)


pkg/reporter/defect_dojo.go
```diff
@@ -27,12 +27,15 @@ type DefectDojoProduct struct {
...
-	Name    string
-	Version string
+	Name           string
+	Version        string
+	InformationURI string
 }
 
 type DefectDojoReporterConfig struct {
 	Tool               DefectDojoToolMetadata
+	IncludeVulns       bool
```
abhisek, line 37: I think `IncludeVulns` and `IncludeMalware` should be enabled by default. Do you see any backward compatibility issue if we enable by default?

pkg/reporter/sarif_builder.go
```diff
@@ -166,3 +171,54 @@ func (b *sarifBuilder) buildFilterResultMessageMarkdown(event *analyzer.Analyzer
...
+		return
+	}
+
+	for _, vuln := range vulns {
+		vulnId := utils.SafelyGetValue(vuln.Id)
+		if _, ok := b.violationsCache[vulnId]; ok {
+			continue
+		}
+
+		b.violationsCache[vulnId] = true
```
abhisek, line 191: `violationsCache` seems like doesn't fit here. Can we rename the `violationsCache` to only `entityCache` or something that is generic enough to hold both violation and vulns? Alternatively we can introduce `vulnCache` separately to avoid ambiguity and ID conflict resulting in subtle bugs

pkg/reporter/defect_dojo.go
KunalSin9h: yes, there is no need for `--report-sarif-vulns --report-sarif-malware` they should be default. For example in `--report-json` we have both policy violations and vulns and malware everyting.

OmkarPh: For defect dojo, yes we can enable by default, maybe i'll just refactor it to have options `SkipVulns` & `SkipMalware` 

However, for plain sarif reports, we need to enable this only when flags are set right, as mentioned in #434 ?

pkg/reporter/sarif_builder.go
OmkarPh: yep, makes sense

pkg/reporter/defect_dojo.go
KunalSin9h: hey @OmkarPh can can't file where its mentioned in #434 can you point me there.

scan.go
```diff
@@ -479,9 +486,12 @@ func internalStartScan() error {
 		engagementName := fmt.Sprintf("vet-report-%s", time.Now().Format("2006-01-02"))
 		rp, err := reporter.NewDefectDojoReporter(reporter.DefectDojoReporterConfig{
 			Tool: reporter.DefectDojoToolMetadata{
-				Name:    "vet",
-				Version: version,
+				Name:           "vet",
+				Version:        version,
+				InformationURI: "https://github.com/safedep/vet",
```
KunalSin9h, line 491: use const variable as mentioned as oppose to string liternal for information uri.

scan.go
```diff
@@ -458,10 +462,13 @@ func internalStartScan() error {
 	if !utils.IsEmptyString(sarifReportPath) {
 		rp, err := reporter.NewSarifReporter(reporter.SarifReporterConfig{
 			Tool: reporter.SarifToolMetadata{
-				Name:    "vet",
-				Version: version,
+				Name:           "vet",
```
KunalSin9h, line 465: we can also put "vet" in `common.go` since this string literal is also used in multiple places.

scan.go
```diff
@@ -479,9 +486,12 @@ func internalStartScan() error {
 		engagementName := fmt.Sprintf("vet-report-%s", time.Now().Format("2006-01-02"))
 		rp, err := reporter.NewDefectDojoReporter(reporter.DefectDojoReporterConfig{
 			Tool: reporter.DefectDojoToolMetadata{
-				Name:    "vet",
-				Version: version,
+				Name:           "vet",
```
KunalSin9h, line 489: we can also put "vet" in common.go since this string literal is also used in multiple places.

pkg/reporter/defect_dojo.go
OmkarPh: my bad, wanted to mention this- https://github.com/safedep/vet/issues/430

pkg/reporter/sarif_builder_test.go
```diff
@@ -149,10 +153,14 @@ func TestSarifReportBuilderMarkdown(t *testing.T) {
...
-	builder, err := newSarifBuilder(sarifBuilderToolMetadata{
-		Name:    "tool-name",
-		Version: "tool-version",
-	})
+	builder, err := newSarifBuilder(
+		sarifBuilderConfig{
+			Tool: sarifBuilderToolMetadata{
+				Name:           "tool-name",
+				Version:        "tool-version",
+				InformationURI: "https://github.com/safedep/vet",
```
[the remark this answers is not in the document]
OmkarPh: agreed, i think we should this for all reporter constructors

pkg/reporter/defect_dojo.go
KunalSin9h: > We can include by default i think since this is an enhancement and not a breaking change.

But @abhisek said to include it by default without flags.

KunalSin9h:
image

There might be case when summary is empty in case of **PYSEC-2024-38**

image

https://osv.dev/vulnerability/PYSEC-2024-38

Can we do something to show some default text like "Found vulnerability in pkg@vertion" in these cases.

pkg/reporter/defect_dojo.go
OmkarPh: i had asked what to do for defect dojo, i think he responded specifically for that

OmkarPh:
> Can we do something to show some default text like "Found vulnerability in pkg@version" in these cases.

Yes, having some text would be great, however I don't think we can correctly say just one version or range in summary, since affected ranges are not always straightforward, `Found vulnerability in pkg` should be good ig

pkg/reporter/sarif_builder.go
```diff
@@ -166,3 +171,54 @@ func (b *sarifBuilder) buildFilterResultMessageMarkdown(event *analyzer.Analyzer
...
+
+		result := sarif.NewRuleResult(vulnId)
+		result.WithLevel("error")
```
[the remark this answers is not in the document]
OmkarPh: ack

scan.go
```diff
@@ -478,10 +490,9 @@ func internalStartScan() error {
 
 		engagementName := fmt.Sprintf("vet-report-%s", time.Now().Format("2006-01-02"))
 		rp, err := reporter.NewDefectDojoReporter(reporter.DefectDojoReporterConfig{
-			Tool: reporter.DefectDojoToolMetadata{
-				Name:    "vet",
-				Version: version,
-			},
+			Tool:               toolMetadata,
+			IncludeVulns:       true,
+			IncludeMalware:     true,
```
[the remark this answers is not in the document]
OmkarPh: intentionally enabled to true

scan.go
```diff
@@ -478,10 +490,9 @@ func internalStartScan() error {
 
 		engagementName := fmt.Sprintf("vet-report-%s", time.Now().Format("2006-01-02"))
 		rp, err := reporter.NewDefectDojoReporter(reporter.DefectDojoReporterConfig{
-			Tool: reporter.DefectDojoToolMetadata{
-				Name:    "vet",
-				Version: version,
-			},
+			Tool:               toolMetadata,
+			IncludeVulns:       true,
```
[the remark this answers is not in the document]
OmkarPh: intentionally enabled to true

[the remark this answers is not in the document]
KunalSin9h: @OmkarPh only for defect dojo right!

abhisek (approved):

KunalSin9h:
@OmkarPh have to done empty messages one

KunalSin9h (approved):

OmkarPh:
> @OmkarPh have to done empty messages one

yep, done

PR #444, opened 2025-04-03
feat(reporter): sync malware analysis report

### LINKED ISSUE #426: Add Support for Publishing Malicious Package Info to Control Tower
`vet` has information about malicious package scan data when run with `--malware` flag. We need to publish this information as part of `--report-sync` to Control Tower.

API spec:
https://github.com/safedep/api/pull/41

Refer to sync reporter on how info is synced to Control Tower:
https://github.com/safedep/vet/blob/main/pkg/reporter/sync.go

arunanshub:
closes #426 

- **build(deps): update dependencies**
- **feat(reporter): sync malware analysis report**


pkg/reporter/sync.go
```diff
@@ -516,6 +516,24 @@ func (s *syncReporter) syncPackage(pkg *models.Package) error {
 		})
 	}
 
+	// Add malware analysis information if available
+	if mar := pkg.GetMalwareAnalysisResult(); mar != nil {
+		req.MaliciousPackageInsight = &controltowerv1.PublishPackageInsightRequest_MaliciousPackageInsight{
+			AnalysisId: mar.AnalysisId,
+			IsMalware:  mar.IsMalware,
+			IsVerified: mar.VerificationRecord != nil && mar.VerificationRecord.GetIsMalware(),
```
abhisek, line 524: `IsVerified` should be independent of if it is malware or not right? We can even verify and mark a package as `NOT MALWARE`. This is particularly useful to deal with false positives. I think we should just check if verification record is available to make a decision on `IsVerified`

pkg/reporter/sync.go
```diff
@@ -516,6 +516,24 @@ func (s *syncReporter) syncPackage(pkg *models.Package) error {
 		})
 	}
 
+	// Add malware analysis information if available
+	if mar := pkg.GetMalwareAnalysisResult(); mar != nil {
```
abhisek, line 520: I know we don't have a test case for this repoter but can you explore a bit if we can have some test cases by mocking the grpc connection that is passed as input to this reporter. It will be good to take this PR as an opportunity to improve the test coverage

But do so without making significant refactoring for this reporter which would increase the risk of breaking it.

abhisek (approved):

PR #450, opened 2025-04-07
Fixed incorrect dependencies provided by Package.GetDependencies

### LINKED ISSUE #433: Incorrect & inconsistent dependencies parsed from package-lock.json
Reproduce with [package-lock.json](https://github.com/user-attachments/files/19487291/package-lock.json)

Test with following code at  https://github.com/safedep/vet/blob/c156f542743e90ee11bc66c17ac770bc57619105/pkg/reporter/summary.go#L147-L151

```golang
deps, _ := pkg.GetDependencies()
fmt.Print("Package -", pkg.GetPackageUrl(), " deps -", len(deps), ": ")
for _, dep := range deps {
	fmt.Print(dep.GetName(), ", ")
}
fmt.Println()
```

Run -
```bash
vet -d -l - scan -M ~/Downloads/package-lock.json
```

OmkarPh:
Fixes #433

### DIFF
--- a/pkg/models/models.go
+++ b/pkg/models/models.go
@@ -460,16 +460,14 @@ func (p *Package) GetDependencies() ([]*Package, error) {
 
 	nodes := graph.GetNodes()
 	for _, node := range nodes {
-		if node.Root {
-			continue
-		}
-
 		if node.Data == nil {
 			continue
 		}
 
-		if p.GetName() != node.Data.GetName() &&
-			p.GetVersion() != node.Data.GetVersion() &&
+		// Multiple versions of the same package can exist in the graph,
+		// find the one that matches the package name, version & ecosystem
+		if p.GetName() != node.Data.GetName() ||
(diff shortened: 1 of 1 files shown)


pkg/models/models.go
```diff
@@ -460,16 +460,14 @@ func (p *Package) GetDependencies() ([]*Package, error) {
...
-
 		if node.Data == nil {
 			continue
 		}
 
-		if p.GetName() != node.Data.GetName() &&
-			p.GetVersion() != node.Data.GetVersion() &&
+		// Multiple versions of the same package can exist in the graph,
+		// find the one that matches the package name, version & ecosystem
+		if p.GetName() != node.Data.GetName() ||
```
abhisek, line 469: @OmkarPh Can we have a test case reproducing the issue and validating the fix?

abhisek (approved):

PR #458, opened 2025-04-10
feat: github repos version resolution

### LINKED ISSUE #455: Suport github repo latest version resoltion for malware inspect
do after https://github.com/safedep/dry/pull/17 is merged.

KunalSin9h:
fixes #455 

```bash
./vet inspect malware --purl pkg:/github/actions/checkout
```
out

```bash
./vet inspect malware --purl pkg:/github/actions/checkout@latest
```
new

image


pkg/common/purl/purl.go
```diff
@@ -31,17 +32,26 @@ func ParsePackageUrl(purl string) (*purlResponseWrapper, error) {
 	}
 
 	if instance.Version == "" || instance.Version == "latest" {
+		ui.PrintMsg("Resolving latest version")
```
abhisek, line 35: Don't use `ui.PrintMsg` in library code. There is a specific reason why UI is in `internal` package. We want it to be internal to `vet` and specific to `vet`. Lib code is reusable, in `vet` and outside `vet`. Use `logger.Debugf/Infof` if required. `UI` functions should be used only in `cmd` files to ensure every part of the system can be re-used outside `vet` as well.

Imagine we build a `vet-server` that uses `vet` as a Go module.

pkg/common/purl/purl.go
```diff
@@ -31,17 +32,26 @@ func ParsePackageUrl(purl string) (*purlResponseWrapper, error) {
...
+			packageName = fmt.Sprintf("%s/%s", instance.Namespace, instance.Name)
+		}
+
+		version, err := common.ResolvePackageLatestVersion(packageName, ecosystem)
 		if err != nil {
 			return nil, err
 		}
 
 		instance.Version = version
+		ui.PrintSuccess("Resolved latest version: %s", version)
```
abhisek, line 54:
```suggestion
```

pkg/common/purl/purl.go
```diff
@@ -6,6 +6,7 @@ import (
 	packagev1 "buf.build/gen/go/safedep/api/protocolbuffers/go/safedep/messages/package/v1"
 	"github.com/google/osv-scanner/pkg/lockfile"
 	"github.com/package-url/packageurl-go"
+	"github.com/safedep/vet/internal/ui"
```
abhisek, line 9:
```suggestion
```

pkg/common/purl/purl.go
abhisek: Also it seems to me we should not resolve version here. PURL parser is only for parsing. Such low level library should not automatically handle anything. Its more of "application logic". This is why we handle the case of GitHub Action commit sha resolution at `cmd` level because thats app logic.

Can we avoid doing version resolution or any "automatic" stuff in `pkg/common`. To support `vet scan --purl`, look at PURL reader in `pkg/readers`

KunalSin9h:
@abhisek resolved.

abhisek (approved):

PR #487, opened 2025-05-20
Fix/ Bug in Resolving Package Version in Maven POM with Dependency Management

### LINKED ISSUE #425: Fix Bug in Resolving Package Version in Maven POM with Dependency Management
Just starting to evaluate vet and I have encountered problem that may result in not valid report.

I start scan with:
```
vet scan -D /tmp/test-test/transactionservice --report-json report.json
```

and my results contains something like this:
```
  "packages": [
    {
      "package": {
        "ecosystem": "Maven",
        "name": "io.swagger.core.v3:swagger-annotations",
        "version": "0"
      },
      "manifests": [
        "e7de49b70dae8f0e"
      ]
    },
    {
      "package": {
        "ecosystem": "Maven",
        "name": "org.projectlombok:lombok",
        "version": "0"
      },
```

is there a way to fix version 0 problem and instead detect proper one?

KunalSin9h:
fixes #425 

Screenshots:

For `parent bom` resolution

Screenshot from 2025-05-20 14-45-41

For `simple pom.xml`

image

Result of `--resport-json`

[report.json](https://github.com/user-attachments/files/20340246/rep.json)

image


pkg/parser/pomxml.go
```diff
@@ -0,0 +1,105 @@
...
+// We use osc-scalibr's java/pomxmlnet (with Net, or Network) to fetch dependency from registry.
+func parseMavenPomXmlFile(lockfilePath string, _ *ParserConfig) (*models.PackageManifest, error) {
+	// Disable osv-scalibr's native logging.
+	log.SetLogger(silentLogger{})
+
+	// Java/PomXMLNet extractor
+	ext, err := el.ExtractorsFromNames([]string{"java/pomxmlnet"})
+	if err != nil {
+		logger.Errorf("Failed to create java/pomxmlnet extractor form osv-scalibr: %v", err)
+		return nil, fmt.Errorf("failed to create java/pomxmlnet extractor: %v", err.Error())
```
abhisek, line 27:
```suggestion
		return nil, fmt.Errorf("failed to create java/pomxmlnet extractor: %w", err)
```

pkg/parser/pomxml.go
```diff
@@ -0,0 +1,105 @@
...
+
+	// ScanConfig
+	config := &scalibr.ScanConfig{
+		ScanRoots:            scanRoots,
+		FilesystemExtractors: ext,
+		Capabilities:         capability,
+		PathsToExtract:       []string{lockfilePath},
+	}
+
+	result := scalibr.New().Scan(context.Background(), config)
```
abhisek, line 58: Nothing much you can do here but just a note: We made a mistake in `vet` design ie. not having a top level context which we pass down to all scanner, analyser, parser etc. At some point, we need to refactor and have a top level context passed down to parsers.

pkg/parser/pomxml.go
```diff
@@ -0,0 +1,105 @@
...
+
+	if result.Status.Status != plugin.ScanStatusSucceeded {
+		logger.Warnf("osv-scalibr scan did not performed scan with success")
+		return nil, fmt.Errorf("osv-scalibr scan did not performed scan with success: Status %v", result.Status.Status)
```
abhisek, line 62: `%v` is not a good idea for `Errorf`. Either use `%w` to wrap underlying error or `%s` to embed error string

pkg/parser/pomxml.go
```diff
@@ -0,0 +1,105 @@
...
+		}
+		manifest.AddPackage(modelPackage)
+	}
+
+	return manifest, nil
+}
+
+// scanRoots function returns the default scan root required for osv-scalibr
+// Default is `/`
+func scanRoots() ([]*scalibrfs.ScanRoot, error) {
```
abhisek, line 81: `scanRoots` and `silentLogger` seems like common code for intefacing with SCAlibr. May be we can have a `scalibr.go` in `parser` package that contains this code. This and other parsers can just re-use this code beyond only `pom.xml` parser

pkg/parser/pomxml.go
```diff
@@ -0,0 +1,105 @@
+package parser
+
+import (
+	"context"
+	"fmt"
+	scalibr "github.com/google/osv-scalibr"
+	"github.com/google/osv-scalibr/binary/platform"
+	el "github.com/google/osv-scalibr/extractor/filesystem/list"
+	scalibrfs "github.com/google/osv-scalibr/fs"
+	"github.com/google/osv-scalibr/log"
```
abhisek, line 10: Can we name this `scalibrlog`. Just to avoid ambiguity?

PR #489, opened 2025-05-23
feat: container scanning

### LINKED ISSUE #342: Container Vulnerability Scanning with OSV-SCALIBR
Reference: 
- https://github.com/google/osv-scalibr
- https://www.notion.so/safedep-inc/Analysis-of-Container-Scanning-Support-for-vet-1ef61d70b236805b9b13d047a283454a

KunalSin9h:
fixes #342 

carbon


pkg/readers/container_image_reader.go
```diff
@@ -0,0 +1,148 @@
...
+	Image string
+}
+
+type containerImageReader struct {
+	config *ContainerImageReaderConfig
+}
+
+var _ PackageManifestReader = &containerImageReader{}
+
+func NewContainerImageReader(config *ContainerImageReaderConfig) (*containerImageReader, error) {
```
abhisek, line 29: We should separate the config and the image target. They are different. `ImageTarget` such as `node:latest` is the target of scan / reading while config defines "how" it is read. I suggest take 2 param as input - `config` and `imageRef`. Fail fast if the `imageRef` is invalid in `NewContainerImageReader`

pkg/readers/container_image_reader_test.go
```diff
@@ -0,0 +1,31 @@
+package readers
+
+import (
+	"github.com/safedep/vet/pkg/models"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestContainerImageReaderEnumManifest(t *testing.T) {
```
abhisek, line 9: This test will significantly slow down development because it will pull the image. We should move container image reader tests to E2E that runs in CI. Have simple validation tests here that are fast.

In future, we need to see if we can mock the results to test other cases.

pkg/readers/container_image_reader.go
```diff
@@ -0,0 +1,148 @@
...
+		config: config,
+	}, nil
+}
+
+func (c containerImageReader) Name() string {
+	return "Container Image Reader"
+}
+
+func (c containerImageReader) ApplicationName() (string, error) {
+	return defaultApplicationName, nil
```
abhisek, line 40: We should return an PURL here because it can represent the source of the container image. 

See: https://github.com/package-url/purl-spec/blob/main/PURL-TYPES.rst#oci

pkg/readers/container_image_reader.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+func (c containerImageReader) EnumManifests(handler func(*models.PackageManifest, PackageReader) error) error {
+	image, err := c.getScalibrContainerImage()
+	if err != nil {
+		logger.Errorf("failed to get Scalibr container image: %s", err)
+		return fmt.Errorf("failed to get Scalibr container image: %s", err)
```
abhisek, line 47:
```suggestion
		return fmt.Errorf("failed to get container image: %s", err)
```

`Scalibr` is internal library. It doesn't make sense to expose it as a error message to user. Logging is fine.

pkg/readers/container_image_reader.go
```diff
@@ -0,0 +1,148 @@
...
+	if err != nil {
+		logger.Errorf("failed to get scan config: %s", err)
+		return fmt.Errorf("failed to get Scalibr scan config: %s", err)
+	}
+
+	// Scan Container
+	result, err := scalibr.New().ScanContainer(context.Background(), image, scanConfig)
+	if err != nil {
+		logger.Errorf("failed to perform osv-scalibr scan: %s", err)
+		return fmt.Errorf("failed to perform osv-scalibr scan: %s", err)
```
abhisek, line 60:
```suggestion
		return fmt.Errorf("failed to perform container scan: %s", err)
```

pkg/readers/container_image_reader.go
```diff
@@ -0,0 +1,148 @@
...
+
+	manifests := make(map[string]*models.PackageManifest)
+
+	for _, pkg := range result.Inventory.Packages {
+		pkgPurl := converter.ToPURL(pkg)
+		packageSet[pkgPurl.String()] = pkg
+	}
+
+	for _, pkg := range packageSet {
+		if _, ok := manifests[pkg.Ecosystem()]; !ok {
```
abhisek, line 75: We are creating manifest by ecosystem. This seems a bit ambiguous. Is there a way to know the "file" from which Scalibr extracted the package? If so we should use the name or ref or PURL to that file (whatever is available) to group the manifests along with the ecosystem

pkg/readers/container_image_reader.go
```diff
@@ -0,0 +1,148 @@
...
+			manifests[pkg.Ecosystem()] = models.NewPackageManifestFromLocal("", pkg.Ecosystem())
```
abhisek, line 76: We should use PURL here using `NewPackageManifestFromPurl`. The PURL should be referencing the container image being scanned.

pkg/readers/container_image_reader.go
```diff
@@ -0,0 +1,148 @@
...
+
+		manifests[pkg.Ecosystem()].AddPackage(pkgPackage)
+	}
+
+	// TODO: Some Ecosystem is very bad, like for alpine packages the ecosystem is Alpine2.25,
+	for _, manifest := range manifests {
+		err = handler(manifest, NewManifestModelReader(manifest))
+		if err != nil {
+			logger.Errorf("failed to process manifest: %s", err)
+			return fmt.Errorf("failed to process manifest: %s", err)
```
abhisek, line 93:
```suggestion
			continue
```

Don't fail. Only log error.

pkg/readers/container_image_reader.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	// Get default scalibr capabilities
+	capability := parser.ScalibrDefaultCapabilities()
+
+	// From Docs: RunningSystem is "Whether the scanner is scanning the real running system it's on"
+	// For Remote Images (Current State), a running system should be false
+	// We're scanning a Linux container image whose filesystem is mounted to the host's disk.
+	// Ref: https://github.com/google/osv-scalibr/blob/a349e505ba1f0bba00c32d3f2df59807939b3db5/binary/cli/cli.go#L574
+	capability.RunningSystem = false
```
abhisek, line 130: See the problem with coupling with `parser` here. Tomorrow if a dev working on parser introduces a new default config then this module will break or misbehave silently. We should never couple config unless they belong to the same domain.

pkg/readers/container_image_reader.go
KunalSin9h: yes, pkg.Locations returns `[]string` of all files from which package is found.

test/e2e_container_image_reader_test.go
```diff
@@ -0,0 +1,108 @@
...
+				})
+
+				assert.NoError(t, err)
+			}
+		})
+	}
+
+}
+
+func TestContainerImageReaderApplicationName(t *testing.T) {
```
abhisek, line 78: This is not E2E. It should be in `container_image_reader_test.go`

We are only keeping end-end tests in `test` package. Unit tests stay with the original code as per Go's convention

abhisek (approved):

Issue #492, opened 2025-05-27
Convert local docker image to tar then scan tar folder.


KunalSin9h:
```bash
vet scan --image myapp:dev
```

This should first check if the image exists using docker client and then convert it into tar using 
```bash
docker save myapp:dev -o myapp.tar
```

and run this tar scanning. 

After the scan is finished, it should clean up ie. remove the temporary tar file irrespective of success or failure

PR #497, opened 2025-05-28
Support for local docker and tar images in container scanning. 

### LINKED ISSUE #491: Allow local `tar` file scanning
```bash
vet scan --image /path/to/node20.tar
```

This should detection path, and should use scalibr's tar scanning feature.

KunalSin9h:
fixes #491 #492


test/scenarios/scenario-12-image-scan-local-docker-image.sh
```diff
@@ -0,0 +1,11 @@
...
+
+set -x
+
+# Scenario for Container Scanning from local docker images
+
+# Pull nats:2.10 image
+docker pull nats:2.10
+
+$E2E_VET_BINARY \
+  scan --image nats:2.10
```
abhisek, line 11: How do we know that it is not pulling?

pkg/readers/container_image_resolve_workflow.go
```diff
@@ -0,0 +1,140 @@
...
+
+	if err := tempTarFile.Close(); err != nil {
+		return "", err
+	}
+
+	// from docs: it is safe to call Name after Close
+	return tempTarFile.Name(), nil
+}
+
+func checkPathExists(path string) (bool, error) {
```
abhisek, line 130: We should have a unit test for this function. Also why is this outside the reader and at the package level? Do we need this function across different readers? If not, I think it should be within the `containerImageReader` struct

test/scenarios/scenario-12-image-scan-local-docker-image.sh
KunalSin9h: image

abhisek (approved):

PR #504, opened 2025-06-02
feat: adds support for go.mod graph parser

slayer321:
fixes https://github.com/safedep/vet/issues/420

Below is the generated graph for sample `go.mod` file in `parser/fixtures/go` folder
#### Note

This PR just adds the scalibr pkg for parsing go.mod file and the graph generated can be incorrect.
In future we need to take a deeper look at how `go mod graph` is implemented .. and we can implement in future iteration

image


abhisek:
@slayer321 CI test seems to be failing

pkg/parser/gomod_test.go
```diff
@@ -0,0 +1,26 @@
...
+	"gopkg.in/inf.v0",         // indirect
+}
+
+func Test_GomodParser_Simple(t *testing.T) {
+	manifest, err := parseGoModFile("./fixtures/go/go.mod", &ParserConfig{})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	assert.Equal(t, 4, len(manifest.Packages))
```
abhisek: We need to retain the dependency relationship. In this example: 3 are direct and only 1 indirect.

pkg/parser/gomod.go
```diff
@@ -0,0 +1,34 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to read the file: %w", err)
+	}
+
+	file, err := modfile.ParseLax(lockfilePath, data, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse go.mod file: %w", err)
+	}
+	manifest := models.NewPackageManifestFromLocal(lockfilePath, models.EcosystemGo)
+
```
abhisek, line 39: We need to build the dependency graph. See example:
https://github.com/safedep/vet/blob/main/pkg/parser/npm_graph.go#L113

Thats the primary requirement of issue #420 

Without dependency graph we are just replacing the existing parser which works fine but does not differentiates between direct and indirect dependencies.

abhisek (changes requested):

KunalSin9h (dismissed):

pkg/parser/gomod.go
KunalSin9h: @slayer321 @abhisek there is `indirect` property in result of this parser, we can find which packages are indirect. 

image

abhisek: @slayer321 I don't see this comment resolved. I still see a list of packages being parsed from Go.mod and not the dependency graph being constructed. Try `go mod graph`. It will generate the dependency graph.

A dependency graph captures the relationship between the different packages. See example here:
https://github.com/safedep/vet/blob/main/pkg/parser/npm_graph.go#L113

Without a dependency graph, `--report-graph` will not work

You need to assert / verify this graph as part of the parser test cases.

slayer321:
@abhisek below are the things I have found about adding dependency graph for go.mod

1. Looked into the internals of  osv-scalibr module for using [gomod extractor](https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/language/golang/gomod/gomod.go) ..it quite same as using the `golang.org/x/mod/modfile` as we are doing it currently in our codebase. The only difference I found was they support for some of the [older version of Go i.e below 1.17](https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/language/golang/gomod/gomod.go#L91) where the get the dependency from go.sum . Though I just gives us the direct and indirect dependency from go.mod file and not the transitive one. 

2. I was looking into the working to `go mod graph` command . As when we run this command it gives us all the dependency including transitive one. After checking some the [internals](https://github.com/golang/go/blob/master/src/cmd/go/internal/modcmd/graph.go#L71) of it I found that they try to load all the transitive dependency from the go.mod file of the indirect dependency (I'm not 100% sure but after having a quick overview that's what I understand).
3. I was trying to find if there are any ways we can generate the graph ourself using `go.mod` file .. I found that there are two tools that is trying to do somethings similar 
    1.  [depstat](https://github.com/kubernetes-sigs/depstat/tree/main) : It's a cli tool which has a subcommand `graph` which generate the `.dot` file. After checking there internals I found that they still uses `go mod graph` command and then generate dependency graph out of it
    2. [modview](https://github.com/bayraktugrul/modview/): this tool also uses `go mod graph` command and generate the dependency graph

After doing all the research till now what I understand is there is no direct way to generate the dependency graph from `go.mod` file. And I think we are not looking to have a dependency on  `go` binary as it might break in container or environment where Go is not installed.

Please let me know what are your thoughts on this.

abhisek:
@slayer321 It seems to me we should just use scalibr for `go.mod` parsing while enforcing the checks for Direct/Indirect based on our config (IncludeTransitive). The reason being, scalibr handles additional cases for older version of go.mod which we need as well because `vet` users may be scanning Go projects that uses old tool chains.

We should not be duplicating the same approach that scalibr is using. Instead, if possible we should be contributing to scalibr by extending its functionality.

what do you think?

slayer321:
@abhisek Yes, we can use scalibr for parsing `go.mod` file(I already tried it yesterday.. as we discuss by directly calling [Extractor](https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/language/golang/gomod/gomod.go#L78)) . But it will just parse direct/indirect dependency that we have in `go.mod` and not the `Transitive` dependency . 

As for now there seems to be not directly way to build the dependency graph for Transitive dependency from `go.mod` other then using `go mod graph` command. 

So for now If we just want to have parsing of  direct/indirect dependency supported and not the graph .. we can use scalibr and implement it as we are already doing for `pom.xml` file.

abhisek:
@slayer321 Sounds good. Lets go ahead as the first iteration with scalibr.

pkg/parser/gomod.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/google/osv-scalibr/extractor/filesystem"
+	"github.com/google/osv-scalibr/extractor/filesystem/language/golang/gomod"
+	"github.com/google/osv-scalibr/fs"
+	"github.com/safedep/vet/pkg/models"
+)
+
+// parseGoModFile parses the go.mod file of the  project.
+func parseGoModFile(lockfilePath string, config *ParserConfig) (*models.PackageManifest, error) {
+	goModExtractor := gomod.New()
```
abhisek: `goModExtractor` initialization should be closer to the code where `goModExtractor` is used. Also this initialization is unnecessary if `os.Open` fails. We should always fail fast.

See Clean Code guidelines:
https://gist.github.com/wojteklu/73c6914cc446146b8b533c0988cf8d29#source-code-structure

Originally referred in book: Clean Code

pkg/parser/gomod.go
```diff
@@ -0,0 +1,47 @@
...
+
+	inventory, err := goModExtractor.Extract(context.Background(), inputConfig)
+	if err != nil {
+		return nil, fmt.Errorf("failed to extract packages: %s", err)
+	}
+
+	manifest := models.NewPackageManifestFromLocal(lockfilePath, models.EcosystemGo)
+
+	for _, pkg := range inventory.Packages {
+		pkgDetails := models.NewPackageDetail(models.EcosystemGo, pkg.Name, pkg.Version)
```
abhisek, line 41: As discussed earlier (in Slack AFAIR), we should include a `ParserConfig` called `IncludeTransitiveDependencies`. This parser should honour that ie. include indirect Go dependencies only if `IncludeTransitiveDependencies` is true.

In `parser.go`, we can consider `IncludeTransitiveDependencies` as true by default.
https://github.com/safedep/vet/blob/main/pkg/parser/parser.go#L272

pkg/parser/gomod_test.go
```diff
@@ -0,0 +1,27 @@
...
+var goDeps = []string{
+	"stdlib",                  // Direct
+	"connectrpc.com/connect",  // Direct
+	"github.com/anchore/syft", // Direct
+	"github.com/gocql/gocql",  // Direct
+	"gopkg.in/inf.v0",         // indirect
+}
+
+func TestGomodParserSimple(t *testing.T) {
+	manifest, err := parseGoModFile("./fixtures/go/go.mod", &ParserConfig{})
```
abhisek: Test for both the cases ie. `IncludeTransitiveDependencies=true` and `IncludeTransitiveDependencies=false`

slayer321: Yes, earlier I [added it](https://github.com/safedep/vet/pull/504/commits/fc086cc47e27773196850a57468058a3db4f8914#diff-6c9d04f16ac3f863642955648fecf7babf69933c0365f0b04a8458d4892c7c40R28) but have to remove it as now we are using scalibr package and it don't expose that `Indirect` field.

pkg/parser/gomod.go
slayer321: Yes, we discuss on that . But now as we are using scalibr package we don't have access to that field as scalibr only expose basic [details](https://github.com/google/osv-scalibr/blob/main/extractor/filesystem/language/golang/gomod/gomod.go#L128) . If you see the previous commit that I have where we where directly using the golang's [modfile](golang.org/x/mod/modfile) pkg .. there we had access to that field so the [checks where added](https://github.com/safedep/vet/pull/504/commits/fc086cc47e27773196850a57468058a3db4f8914#diff-c4db7e394223f198b84d5801cebb3d8b0865a364f1165ef1b0a169a91b1f4765R24) .

pkg/parser/gomod_test.go
```diff
@@ -0,0 +1,39 @@
...
+		t.Fatal(err)
+	}
+
+	assert.Equal(t, 5, len(manifest.Packages))
+	for _, pkg := range manifest.Packages {
+		assert.Contains(t, goDeps, pkg.Name)
+	}
+}
+
+func TestGomodParserToExcludeTransitiveDependencies(t *testing.T) {
```
abhisek: Unnecessary test function. Should combine the 2 test functions following Go's table driven testing convention into a single function but multiple test cases.

https://go.dev/wiki/TableDrivenTests

Issue #398, opened 2025-03-18
Cloud Session Refresh using Refresh Token


KunalSin9h:
## Background

`vet cloud` commands interfact with SafeDep Cloud.
https://docs.safedep.io/cloud

The easiest way to get started is `vet cloud quickstart`. This does all the setup and finally obtains an access token and refresh token using which we can execute commands such as:

`vet cloud key list`, `vet cloud whoami` etc.

Once `vet cloud` auth is done, keys are stored in `~/.safedep/vet-auth.yml` which is accessible to `vet` using its `auth` package functions.

Even though we store the refresh token, we do not leverage it to refresh the session after the access token (JWT) has timed out. This gives a poor user experience. The user have to do `vet cloud login` again to fetch an access token.

## Requirements

Add support for 

- Check if access token is expired (may be by calling whoami API)
- Leverage refresh token to fetch and store a new set of access and refresh token
- Continue using new tokens in any `vet cloud` command
- Fail if its not possible to refresh

As a user, the experience should be transparent. For example, I come back after 7 days and run:

`vet cloud key list`

The system should intelligently make a decision based on `cloud_access_token_updated_at` if it should check for expired access token. It should NOT check always otherwise all `vet cloud` commands will be slowed down.

If the access token is expired, executed the refresh token flow to update the tokens and then continue the existing flow of command execution

abhisek:
@Sahilb315 Please take this in your backlog. You can refer the existing PR and the comments for reference but I suggest you create a PR from scratch.

PR #507, opened 2025-06-04
fix: cloud session refresh using refresh token

slayer321:
fixes https://github.com/safedep/vet/issues/398

two conditions are check before getting new access token
1. If the last updated time is less then 10 hrs we don't refresh (this is done so that we don't add any extra check if refresh is not rquired)
2. If the current time in more than the expired time mentioned in the JWT token. We need to refresh the access token.


cmd/cloud/key.go
```diff
@@ -83,7 +83,11 @@ func newListKeyCommand() *cobra.Command {
 		Use:   "list",
 		Short: "List API keys",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			err := executeListKeys()
+			err := getNewAccessTokenUsingRefreshTokenIfCurrentIsExpired()
```
abhisek: Looks like `getNewAccessTokenUsingRefreshTokenIfCurrentIsExpired` is actually mutating an internal state ie. updating access token. The name is mislead because `getXXX` usually implies read operation.

May be something like `refreshAccessToken` is more intention revealing. The logic of caching / checking if current one is expired can be internal to this function.

cmd/cloud/key.go
```diff
@@ -83,7 +83,11 @@ func newListKeyCommand() *cobra.Command {
...
+			if err != nil {
+				logger.Errorf("Failed to refresh access token: %v", err)
```
abhisek: We should use `ui.PrintError` here to show the error to the user because this error is a possibly fatal error for the command.

cmd/cloud/ping.go
```diff
@@ -15,7 +15,11 @@ func newPingCommand() *cobra.Command {
 		Use:   "ping",
 		Short: "Ping the control plane to check authentication and connectivity",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			err := pingControlPlane()
+			err := getNewAccessTokenUsingRefreshTokenIfCurrentIsExpired()
```
abhisek: I think we are duplicating code and violating DRY principle. Look at Cobra pre-run. You can register some function to be executed before any `vet cloud XXX` command is executed. For that I think you have to use `PersistentPreRun` on the `cmd` for cloud.

cmd/cloud/utils.go
```diff
@@ -0,0 +1,143 @@
+package cloud
```
abhisek: I think this entire code should do to the `internal/auth` package because its related to `auth` and may not be bounded only to the `cloud` command. May be used elsewhere in future.

From a domain perspective, this seems like an operation within the `auth` domain. So I feel its better to put this code in the `auth` package

cmd/cloud/utils.go
```diff
@@ -0,0 +1,143 @@
...
+
+type TokenResponse struct {
+	AccessToken string `json:"access_token"`
+	ExpireIn    int    `json:"expires_in"`
+	Scope       string `json:"scope"`
+	IDToken     string `json:"id_token"`
+	Tokentype   string `json:"token_type"`
+}
+
+func loadConfiguration() (auth.Config, error) {
```
abhisek: Do not duplicate code. Fundamental rule is not to violate DRY. All of these are part of the `internal/auth` package

cmd/cloud/utils.go
```diff
@@ -0,0 +1,143 @@
...
+	return os.WriteFile(path, data, 0o600)
+}
+
+func getNewAccessTokenAndPersistIt(config auth.Config) error {
+	data := url.Values{}
+	data.Set("grant_type", "refresh_token")
+	data.Set("refresh_token", config.CloudRefreshToken)
+	data.Set("client_id", auth.CloudIdentityServiceClientId())
+
+	req, _ := http.NewRequest("POST", auth.CloudIdentityServiceTokenUrl(), strings.NewReader(data.Encode()))
```
abhisek: Try using the `cli/oauth` package to achieve this instead of making manual HTTP call. See `cmd/cloud/quickstart.go`

cmd/cloud/utils.go
```diff
@@ -0,0 +1,143 @@
...
+
+	return persistConfiguration(config)
+}
+
+// checkIfNewAccessTokenRequired does two level of check before trying to get the new access token
+//  1. Using CloudAccessTokenUpdatedAt field it check if the last time it was updated
+//     is less than equal to 10 hrs or not(Here 10hrs was taken because it was observer that most of the JWT token had similar expiry time)
+//  2. It will parse the JWT token to get the expiry time and then check for it expiry
+func checkIfNewAccessTokenRequired(config auth.Config) (bool, error) {
+	if time.Since(config.CloudAccessTokenUpdatedAt) <= 10*time.Hour {
```
abhisek: Whats the rationale of using 10 hour as the window? Especially since later in the code we are actually checking the JWT `exp`

internal/auth/auth.go
```diff
@@ -291,6 +291,10 @@ func CommunityMode() bool {
 	return false
 }
 
+func HomeRelativeConfigPath() string {
```
abhisek: Avoid exposing this unless required because this is internal to this package

cmd/cloud/utils.go
slayer321: The reason for having this two level of checks was to leverage the `CloudAccessTokenUpdatedAt` field that we store on the user side .. so In this case 
1. Using `CloudAccessTokenUpdatedAt` field is current time is less  than 10 hrs when the last time access token was updated .. we don't go ahead to check the JWT check token .. 
2. If the first check is not true then only we check for JWT token expiry

The reason to add the first 10 hour window check was if it is not expired there will be no need to parse JWT token and check for it expiry . 

But as I was thinking more on this . I believe JWT expiry check is more future proof .. 

So let me know if we want to keep both the checks or just the JWT expiry check ?

cmd/cloud/utils.go
slayer321: I check the `cli/oauth` package and can see that it is either used to `deviceCode` or `webflow` as in our case we don't want user to redirect anywhere to get the new access token. 

So I don't think we can use `cli/oauth` package  here.

cmd/cloud/main.go
```diff
@@ -41,6 +42,10 @@ func NewCloudCommand() *cobra.Command {
 		if tenantDomain != "" {
 			auth.SetRuntimeCloudTenant(tenantDomain)
 		}
+		err := auth.RefreshAccessToken()
```
abhisek: @slayer321 Have you verified the flow? It seems to me this will cause problem because it will run even first time when a user does `vet cloud login` and will fail, show the error and cause a poor user experience. The first time user will likely be confused with the error message.

internal/auth/auth.go
```diff
@@ -344,3 +350,78 @@ func persistConfiguration() error {
 	os.MkdirAll(filepath.Dir(path), os.ModePerm)
 	return os.WriteFile(path, data, 0o600)
 }
+
+type TokenResponse struct {
```
abhisek: Is this required to be public outside the package?

internal/auth/auth.go
```diff
@@ -344,3 +350,78 @@ func persistConfiguration() error {
...
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		body, _ := io.ReadAll(resp.Body)
```
abhisek, line 378: Is this required, especially since we have deferred `resp.Body.Close()`?

internal/auth/auth.go
```diff
@@ -344,3 +350,78 @@ func persistConfiguration() error {
...
+		return fmt.Errorf("refresh failed: %s", string(body))
+	}
+
+	var result TokenResponse
+
+	if err := json.NewDecoder(resp.Body).Decode(&result); err != nil {
+		return err
```
abhisek: Best practice is to add a context and chain error. Example: `fmt.Errorf("failed to decode JWT: %w", err)`

internal/auth/auth.go
```diff
@@ -344,3 +350,78 @@ func persistConfiguration() error {
...
+
+type TokenResponse struct {
+	AccessToken string `json:"access_token"`
+	ExpireIn    int    `json:"expires_in"`
+	Scope       string `json:"scope"`
+	IDToken     string `json:"id_token"`
+	Tokentype   string `json:"token_type"`
+}
+
+func getNewAccessTokenAndPersistIt() error {
```
abhisek: This function violates CQRS. The name indicates `getXX` which is a read operation. But internally it updates global state (config) as well as persist the token. This function has side-effect. Should be named accordingly or split

cmd/cloud/main.go
slayer321: yes, I check the flow. But miss the part what if user running the command for the fist time.. In my case I started with the `quickstart` command.. and while running login command I didn't got error . But yes if there is new user and running the login command it will be poor user experience. 
I'm thinking we can add a check here to skip the `login` and `quickstart` for calling  RefreshAccessToken function.

internal/auth/auth.go
slayer321: No, will make it private.

internal/auth/auth.go
slayer321: I think yes, though we are deferring the `resp.Body.Close()` which will close stream at the end of the function.. but what if we didn't get http statusCode as `StatusOK` ... in that case we just want to read the response body and return the error .

PR #518, opened 2025-06-16
feat: Malicious Packages (OSV) Reporter for Inspect Command

abhisek:
Example usage:

```
./vet inspect malware --purl pkg:/npm/x@y \
--report-osv ~/Work/dev/oss/malicious-packages/ 
```

The path for `--report-osv` should be the root of cloned https://github.com/ossf/malicious-packages


abhisek:
@OmkarPh Looks like the dependency usage evidence related E2E are failing. Can you have a look and see how we can avoid the flakiness of the tests?

https://github.com/safedep/vet/actions/runs/15844105054/job/44662226198?pr=518#step:5:383

PR #535, opened 2025-07-08
feat: Add Support for Agentic Query and Analysis

abhisek:
This PR introduces two features:

1. sqlite3 report generation for vet
2. Agentic query system

It also lays the foundation for building AI agents in vet.

To test, consider the doc:

https://github.com/safedep/vet/blob/feat/vet-query-interaction-agent/docs/agent.md

### DIFF
--- a/agent/memory.go
+++ b/agent/memory.go
@@ -15,7 +15,9 @@ type simpleMemory struct {
 var _ Memory = (*simpleMemory)(nil)
 
 func NewSimpleMemory() (*simpleMemory, error) {
-	return &simpleMemory{}, nil
+	return &simpleMemory{
+		interactions: make([]*schema.Message, 0),
+	}, nil
 }
 
 func (m *simpleMemory) AddInteraction(ctx context.Context, interaction *schema.Message) error {
--- a/agent/memory_test.go
+++ b/agent/memory_test.go
@@ -1,0 +1,279 @@
+package agent
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"testing"
+
+	"github.com/cloudwego/eino/schema"
+	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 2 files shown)


KunalSin9h (commented):
docs missing: `--insights-v2`

agent/mcp.go
```diff
@@ -0,0 +1,126 @@
...
+			return nil, fmt.Errorf("failed to create sse client: %w", err)
+		}
+	} else {
+		cli, err = b.buildStdioClient()
+		if err != nil {
+			return nil, fmt.Errorf("failed to create stdio client: %w", err)
+		}
+	}
+
+	err = cli.Start(ctx)
```
KunalSin9h, line 59: @abhisek eino code docs says , we don't need to run `Start` for `stdio` clients. It happens already. We do need to run this in `SSE` clients. 

```
		// From Docs:
		// NOTICE: NewStdioMCPClient will start the connection automatically. Don't call the Start method manually. This is for backward compatibility
```

can run `cli.Start` on in `if b.config.SseURL` branch.

abhisek: Hmm not a good design. Its bad practice to have to do `if (...)` to operate on a standard interface. `Start(..)` not relevant for `stdio` should be an internal detail and not exposed to me as a user.

docs/agent.md
```diff
@@ -0,0 +1,34 @@
...
+```bash
+vet scan --insights-v2 -M package-lock.json --report-sqlite3 report.db
+```
+
+**Note:** Agents only work with `--insights-v2`
+
+* Start the query agent
+
+```bash
+vet agent query --db report.db
```
KunalSin9h, line 36: @abhisek docs for `--fast` flag.

KunalSin9h (approved):

PR #536, opened 2025-07-09
OpenVSX extensions scanning support

Sahilb315:


Sahilb315:
Just for better naming we could also rename `vscode_ext_reader.go` file to `vsix_ext_reader.go` as its not limited to vscode only. `vsix` being the extension format OpenVSX & VSCode follow.

Should we? @abhisek

KunalSin9h (approved):

abhisek:
> Just for better naming we could also rename `vscode_ext_reader.go` file to `vsix_ext_reader.go` as its not limited to vscode only. `vsix` being the extension format OpenVSX & VSCode follow.
> 
> Should we? @abhisek

@Sahilb315 Yes. Let us go ahead

pkg/readers/vscode_ext_reader.go
```diff
@@ -35,16 +41,24 @@ type vsCodeExtensionList struct {
...
+}
+
 type vscodeExtReader struct {
-	distributionHomeDir map[string]string
+	distributions map[string]distributionInfo
 }
 
 var _ PackageManifestReader = (*vscodeExtReader)(nil)
 
 func NewVSCodeExtReader(distributions []string) (*vscodeExtReader, error) {
```
abhisek: Lets rename this to `NewVSIXExtReader?

pkg/readers/vscode_ext_reader.go
```diff
@@ -56,16 +70,30 @@ func NewVSCodeExtReaderFromDefaultDistributions() (*vscodeExtReader, error) {
 		return nil, fmt.Errorf("failed to get user home directory: %w", err)
 	}
 
-	distributionHomeDir := map[string]string{
```
abhisek, line 59: Lets rename the function to VSIX and remove VSCode specifically?

pkg/readers/vscode_ext_reader.go
Sahilb315: Also in this function currently the ecosystem is hardcoded to `vscode`. Shouldn't we add a check/match the IDE's supporting OpenVSX & assign the correct ecosystem.

Something like - 
```go
        ecosystem := models.EcosystemVSCodeExtensions // default to VSCode
        
        // Check if path matches any known OpenVSX-based IDE patterns
        for pattern, eco := range knownIDEPatterns {
            if strings.Contains(distribution, pattern) {
                ecosystem = eco
                break
            }
        }
```

pkg/readers/vsix_ext_reader.go
```diff
@@ -12,6 +12,12 @@ import (
 
 const (
 	vsCodeExtensionExtensionsFileName = "extensions.json"
+
+	// Default paths for IDE extension directories relative to home directory
+	vsCodeExtPath   = ".vscode/extensions"
```
abhisek: Should we maintain this as a map of editor name and a map of path and ecosystem?

pkg/readers/vsix_ext_reader.go
```diff
@@ -35,37 +41,59 @@ type vsCodeExtensionList struct {
 	Extensions []vsCodeExtension
 }
 
+type distributionInfo struct {
+	FilePath  string // Path to the extensions directory
+	Ecosystem string // Type of extension marketplace (VSCode or OpenVSX)
+}
+
 type vscodeExtReader struct {
```
abhisek: Should we change the name to vsix reader?

abhisek (changes requested):

pkg/readers/vscode_ext_reader.go
abhisek: Makes sense. Now that this reader is generic and not VS Code specific. Let's go ahead and use the suitable ecosystem

KunalSin9h (approved):

PR #539, opened 2025-07-14
Add github actions sync resolver

### LINKED ISSUE #530: Add Sync Reporter Environment Resolver for GitHub
The sync reporter syncs `vet` data with SafeDep Cloud. This reporter need _optional_ input from the environment such as source (GitHub, Gitlab, Git etc), repository URL, trigger etc. The contract is defined as:

https://github.com/safedep/vet/blob/main/pkg/reporter/sync.go#L33

Currently we are using a default resolver which just returns empty values.

We need to add support for different CI resolvers such as GitHub, Gitlab etc. To start with, let us support only GitHub because thats where our primary user base exist.

Add support for GitHub Action environment resolver based on GitHub Actions default environment variables.

We also need to add a factory function such as `NewSyncReporterEnvironmentResolver` that leverages some heuristic to check if we are running in GitHub Actions, if so, return the GHA environment resolver.

We should fallback to the existing default resolver (No Op) if GHA environment is not detected.

Sahilb315:
Fixes #530


pkg/reporter/github_sync_reporter.go
```diff
@@ -0,0 +1,53 @@
+package reporter
```
abhisek: The file should be named resolver instead of reporter to avoid confusion that it is a reporter of it's own. It's actually not because it just helps sync reporter to resolve runtime info

pkg/reporter/github_sync_reporter.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"os"
+
+	controltowerv1pb "buf.build/gen/go/safedep/api/protocolbuffers/go/safedep/messages/controltower/v1"
+	controltowerv1 "buf.build/gen/go/safedep/api/protocolbuffers/go/safedep/services/controltower/v1"
+)
+
+type githubActionResolver struct{}
+
+func GHASyncReporterResolver() SyncReporterEnvResolver {
```
abhisek: Let's name it explicitly as GithubAction... please instead of GHA

pkg/reporter/github_sync_reporter.go
```diff
@@ -0,0 +1,53 @@
...
+	return &githubActionResolver{}
+}
+
```
abhisek: Please add tests. We can easily test by setting environment variables and asserting the resolver to return correct values.

pkg/reporter/sync_test.go
```diff
@@ -111,6 +112,86 @@ func mockDependencyGraph(pkg *models.Package) {
...
+		expectedProjectSource controltowerv1pb.Project_Source
+		expectedProjectUrl    string
+		expectedTrigger       controltowerv1.ToolTrigger
+		expectedGitRef        string
+		expectedGitSha        string
+	}{
+		{
+			name: "should return GitHub Actions values when GITHUB_ACTIONS is set",
+			setupEnv: func() {
+				os.Setenv("GITHUB_ACTIONS", "true")
```
abhisek: Uh oh. Use `t.SetEnv` during test so that the test runner can track and cleanup.

https://pkg.go.dev/testing#B.Setenv

pkg/reporter/sync_test.go
```diff
@@ -111,6 +112,86 @@ func mockDependencyGraph(pkg *models.Package) {
...
+			expectedProjectUrl:    "",
+			expectedGitRef:        "",
+			expectedGitSha:        "",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Setup test environment
+			tt.setupEnv()
```
abhisek: Just pass `t` to `tt.setupEnv` and use that to set the env. Cleanup is automatic once `t.Run(..)` finishes.

pkg/reporter/sync_test.go
```diff
@@ -111,6 +112,86 @@ func mockDependencyGraph(pkg *models.Package) {
...
+
+			// Cleanup after test
+			defer tt.cleanupEnv()
```
abhisek: This is not required when you use test runner's env utils

abhisek:
@Sahilb315 The tests are failing for this.

Sahilb315:
> @Sahilb315 The tests are failing for this.

Ohh.. they are passing locally but failing here because the env variables are set in the CI/CD runner & can access the variables.

@abhisek Fixed it

abhisek (approved):

PR #543, opened 2025-07-18
Support exclusion patterns for `lockfiles` flag

Sahilb315:
Fixes https://github.com/safedep/vet-action/issues/118

### DIFF
--- a/pkg/readers/common.go
+++ b/pkg/readers/common.go
@@ -1,6 +1,8 @@
 package readers
 
 import (
+	"path/filepath"
+
 	"github.com/bmatcuk/doublestar/v4"
 	"github.com/safedep/vet/pkg/common/logger"
 )
@@ -19,15 +21,30 @@ func newPathExclusionMatcher(exclusions []string) *exclusionMatcher {
 
 func (ex *exclusionMatcher) Match(term string) bool {
 	for _, exclusionPattern := range ex.Exclusions {
-		m, err := doublestar.Match(exclusionPattern, term)
-		if err != nil {
-			logger.Warnf("Invalid path pattern: %s: %v", exclusionPattern, err)
-			continue
-		}
-
-		if m {
+		// Try matching in current form first
+		if m, err := doublestar.Match(exclusionPattern, term); err == nil && m {
 			return true
 		}
+
+		// If term is relative and pattern is absolute, convert term to absolute
+		if !filepath.IsAbs(term) && filepath.IsAbs(exclusionPattern) {
+			if abs, err := filepath.Abs(term); err == nil {
+				if m, err := doublestar.Match(exclusionPattern, abs); err == nil && m {
+					return true
+				}
+			}
+		}
+
+		// If term is absolute and pattern is relative, convert pattern to absolute
+		if filepath.IsAbs(term) && !filepath.IsAbs(exclusionPattern) {
+			if abs, err := filepath.Abs(exclusionPattern); err == nil {
+				if m, err := doublestar.Match(abs, term); err == nil && m {
+					return true
+				}
+			}
+		}
+
+		logger.Debugf("No match for pattern '%s' against '%s'", exclusionPattern, term)
 	}
 
 	return false


abhisek:
@KunalSin9h Can you please test and verify this once since this is a critical flow and we don't want it to break anything

KunalSin9h:
@Sahilb315 
image

should `exclude: *` should work? not something practical, but theoretical. 
:green_circle:  After a checking regex manual, i found that `*` is not a valid `regex`. So we can walk away with it. 

## The linux / unix CLI Problem

This thing is not going to impact `vet-action` or `go test environment` since it not cli environment, but when we use `--exclude pom*` in cli, even before we constrct the regex or match the regex , the unix is already completing the file name (as any linux should do, say `rm temp-file*`, i.e if we use `pom.*` our `scanExclude: [pom.xml]`, 

Say i have 2 manifest `package.json` and `pom.xml` in directory, and i am using this command

```bash
vetbuild scan -v --exclude p* --lockfiles pom.xml
```

The execpted behaviour will be that both files `pom.xml` and `package.json` should be excluded since they both should match `p*` regex. 

But what `linux` will do it treat `p*` as its domain problem (its concern) and autofill it with first value, it will set `--exclude` to `package.json` (see image) i am using `fmt.Println(scanExclude)` in `scan.go`. 

image

The actual behavior will be `pom.xml` will be scanned. 

:skull: So what is happening in `linux` / `unix` (not sure about `windows`) we will **NEVER** going to have a `regex` to our `lockfile-reader.go` it will resole it the cli command itself to some file

### The solution

we have to use `' '` in regex so that the `regex` will be send to `lockfile-reader` 

image


@Sahilb315 what you commended (bellow this comment)

image

It as unexpected behavior. in Linux / UNIX cli environment.

Sahilb315:
> @Sahilb315 
> image
> 
> should `exclude: *` should work? not something practical, but theoretical. 

Is this valid though? Shouldn't it be either `./.*` (meaning exclude anything starting from current dir) or `/.*` (exclude from root)

abhisek:
@KunalSin9h Good though and analysis. But this is a known problem / feature wrt. Unix shell. So `'...'` is generally used to avoid shell globbing. So it should be `vet scan -D /a/b --exclude 'p*.xml'`

But thinking about it, it seems to me we should not use RegEx when it comes to matching exclusion. Instead we should treat the exclusion pattern as path pattern. Example: https://pkg.go.dev/path#Match

So that, it matches `p*.xml` which is probably not a valid XML but a valid path pattern. In fact, when I am writing paths or file names, intuitively I like to use conventional path patterns and not regex.

What do you think @Sahilb315 If you agree, lets go ahead and fix this behaviour, which seems like a bug in our existing code to me. Because in `vet scan -h` I see `--exclude` says `Name patterns` but internally we treat it as regex instead of path patterns.

May be we can extract the function for matching `--exclude` into a common function that can be re-used by both lockfile and dir reader so that we can guarantee consistent behaviour. Also having a common function just makes it easy for us to test. Something like:

```
m := newPathExclusionMatcher(exclusions)
if m.Match(path) {
// skip path
}
```

Test for cases like the below and more:

```
p*.xml
x/**/*.yml
x/b/name*.json
```

Sahilb315:
@abhisek Yes, we can go ahead with this approach. While adding exclusions for lockfiles i was also a bit confused for why are we relying on regex for file pattern matching :sweat_smile:

Sahilb315:
@abhisek Fixed the bug. Please review once

pkg/readers/common.go
```diff
@@ -1,3 +1,51 @@
...
+	for _, exclusionPattern := range ex.Exclusions {
+		// Try matching in current form first
+		if m, err := doublestar.Match(exclusionPattern, term); err == nil && m {
+			return true
+		}
+
+		// If term is relative and pattern is absolute, convert term to absolute
+		if !filepath.IsAbs(term) && filepath.IsAbs(exclusionPattern) {
+			if abs, err := filepath.Abs(term); err == nil {
+				if m, err := doublestar.Match(exclusionPattern, abs); err == nil && m {
```
abhisek, line 32: Can we not use Go path.Match instead of a 3rd party package?

abhisek: Ah! I get it why this is required

pkg/readers/dir_reader.go
```diff
@@ -54,6 +53,8 @@ func (p *directoryReader) ApplicationName() (string, error) {
 func (p *directoryReader) EnumManifests(handler func(*models.PackageManifest,
 	PackageReader) error,
 ) error {
+	exclusionMatcher := newPathExclusionMatcher(p.config.Exclusions)
```
abhisek, line 56: nitpick: For a future cleanup, we can move initializing `exclusionMatcher` in the `NewXXX` function only because this is pretty much initialization.

abhisek (approved):

Issue #544, opened 2025-07-21
vet mcp server crashing


KunalSin9h:
I am using vet mcp server on `VsCode`. With stdio transport and command

```bash
vet server mcp
```

Some time, on some tool it fails with these error. 

```ocaml
2025-07-21 14:33:24.034 [info] Discovered 6 tools
2025-07-21 14:33:52.880 [warning] [server stderr] panic: runtime error: invalid memory address or nil pointer dereference
2025-07-21 14:33:52.882 [warning] [server stderr] [signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x22bc07b]
2025-07-21 14:33:52.884 [warning] [server stderr] 
2025-07-21 14:33:52.885 [warning] [server stderr] goroutine 1 [running]:
2025-07-21 14:33:52.887 [warning] [server stderr] github.com/safedep/vet/mcp/tools.(*packageInsightsTool).executeGetPackageLicenseInfo(0xc000963840, {0x3c23498, 0xc000dbc000}, {{{0xc000be00a0, 0xa}, {0x0}}, {{0xc000bf2080, 0x20}, {0x3088fc0, 0xc000dbc090}, ...}})
2025-07-21 14:33:52.889 [warning] [server stderr] 	/home/runner/work/vet/vet/mcp/tools/package_insights.go:132 +0x23b
2025-07-21 14:33:52.892 [warning] [server stderr] github.com/mark3labs/mcp-go/server.(*MCPServer).handleToolCall(0xc000632000, {0x3c23498, 0xc000dbc000}, {0x2ea6d60, 0xc000be0048}, {{{0xc000be00a0, 0xa}, {0x0}}, {{0xc000bf2080, 0x20}, ...}})
2025-07-21 14:33:52.893 [warning] [server stderr] 	/home/runner/go/pkg/mod/github.com/mark3labs/mcp-go@v0.31.0/server/server.go:1016 +0x3e2
2025-07-21 14:33:52.894 [warning] [server stderr] github.com/mark3labs/mcp-go/server.(*MCPServer).HandleMessage(0xc000632000, {0x3c23498?, 0xc000a88030?}, {0xc000dae0e0, 0xe0, 0xe0})
2025-07-21 14:33:52.894 [warning] [server stderr] 	/home/runner/go/pkg/mod/github.com/mark3labs/mcp-go@v0.31.0/server/request_handler.go:305 +0xc5f
[...]

PR #548, opened 2025-07-26
fix(tools): prevent nil pointer panic in GetPackageLicenseInfo handler

RohanMishra315:
Fixes: #544

This commit adds a nil check in the GetPackageLicenseInfo handler to avoid potential panics when license information is unavailable.
It ensures the handler gracefully returns an empty result or appropriate error instead of crashing the application


abhisek:
@RohanMishra315 Thanks for your contribution.

PR #551, opened 2025-07-29
add support for extensions purl

Sahilb315:
Add supports for 
- `openvsx` in inspect cmd: `vet inspect malware --purl pkg:/openvsx/streetsidesoftware.code-spell-checker@4.0.47`
- scanning VS Code & OpenVSX extensions through `vet scan`: `vet scan --purl pkg:/vscode/streetsidesoftware.code-spell-checker@4.0.47`

### DIFF
--- a/pkg/common/purl/purl.go
+++ b/pkg/common/purl/purl.go
@@ -77,6 +77,10 @@ func PurlTypeToEcosystem(purlType string) (lockfile.Ecosystem, error) {
 		"rubygems":              lockfile.BundlerEcosystem,
 		packageurl.TypeGithub:   models.EcosystemGitHubActions,
 		"actions":               models.EcosystemGitHubActions,
+		"vscode":                models.EcosystemVSCodeExtensions,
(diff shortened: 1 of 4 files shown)


pkg/common/purl/purl.go
```diff
@@ -77,6 +77,10 @@ func PurlTypeToEcosystem(purlType string) (lockfile.Ecosystem, error) {
 		"rubygems":              lockfile.BundlerEcosystem,
 		packageurl.TypeGithub:   models.EcosystemGitHubActions,
 		"actions":               models.EcosystemGitHubActions,
+		"vscode":                models.EcosystemVSCodeExtensions,
```
abhisek, line 80: This a technical debt. At some point in future, we need to clean this up and move to using DRY as the source of truth for PURL parsing.

abhisek: @Sahilb315 FYI. Put something in the backlog for this.

PR #557, opened 2025-08-11
Fix SLSA CI Failing

Sahilb315:


Sahilb315:
SLSA allows only using of tags as version instead of any commit hashes.
https://github.com/slsa-framework/slsa-github-generator?tab=readme-ov-file#referencing-slsa-builders-and-generators

image_720

abhisek (approved):

abhisek:
@Sahilb315 LGTM. But this is weird. How come the workflow works with tag and not with hash? Something to explore. 

Update: Never mind. I see the docs that you provided.

PR #565, opened 2025-08-13
add cloud session refresh using refresh token functionality

Sahilb315:


cmd/cloud/main.go
```diff
@@ -45,3 +61,20 @@ func NewCloudCommand() *cobra.Command {
 
 	return cmd
 }
+
+func RequireAccessTokenCheck(cmd *cobra.Command, args []string) error {
```
abhisek: Don't think we need to expose this.


```suggestion
func requireAccessTokenCheck(cmd *cobra.Command, args []string) error {
```

cmd/cloud/main.go
```diff
@@ -45,3 +61,20 @@ func NewCloudCommand() *cobra.Command {
...
+	// Check if token was obtained/refreshed 5 mins ago
```
abhisek, line 66: What is the rationale for 5min? Why not 30min?

internal/auth/auth.go
```diff
@@ -42,6 +48,8 @@ const (
 	cloudIdentityServiceBaseUrl       = "https://auth.safedep.io"
 	cloudIdentityServiceDeviceCodeUrl = "https://auth.safedep.io/oauth/device/code"
 	cloudIdentityServiceTokenUrl      = "https://auth.safedep.io/oauth/token"
+
+	accessTokenExpiryCheckInterval = 5 * time.Minute
```
abhisek, line 55: Why 5min?

internal/auth/auth.go
```diff
@@ -344,3 +352,68 @@ func persistConfiguration() error {
...
+	exp, ok := claims["exp"].(float64)
+	if !ok {
+		return false, fmt.Errorf("invalid jwt token")
+	}
+
+	return time.Now().Unix() > int64(exp), nil
+}
+
+func ShouldCheckAccessTokenExpiry() bool {
+	if globalConfig == nil {
```
abhisek: The function `ShouldCheckAccessTokenExpiry` seems like it is meant only to make a decision and not to update global state. This is an unexpected side-effect which should be avoided.

If globalConfig is nil, just return `true` so that access token expiry check goes through. No need to set `globalConfig`

internal/auth/auth.go
```diff
@@ -344,3 +352,68 @@ func persistConfiguration() error {
...
+	}
+
+	globalConfig.CloudAccessToken = response.Token
+	globalConfig.CloudRefreshToken = response.RefreshToken
+
+	return persistConfiguration()
+}
+
+func IsAccessTokenExpired() (bool, error) {
+	if globalConfig == nil {
```
abhisek: This is a hidden side effect. This function should not modify global state. Just return true if `globalConfig` is nil. We have `loadConfiguration` set on `init(..)` which should load global config if available. If not, we should check for refresh token anyway or fail.

internal/auth/auth.go
```diff
@@ -344,3 +352,68 @@ func persistConfiguration() error {
...
+		c := DefaultConfig()
+		globalConfig = &c
+	}
+
+	claims := jwt.MapClaims{}
+	_, _, err := jwt.NewParser().ParseUnverified(globalConfig.CloudAccessToken, claims)
```
abhisek, line 412: Check if `globalConfig.CloudAccessToken` is empty and fail fast (return true)

internal/auth/auth.go
```diff
@@ -344,3 +352,68 @@ func persistConfiguration() error {
...
+
+	exp, ok := claims["exp"].(float64)
+	if !ok {
+		return false, fmt.Errorf("invalid jwt token")
+	}
+
+	return time.Now().Unix() > int64(exp), nil
+}
+
+func ShouldCheckAccessTokenExpiry() bool {
```
abhisek, line 425: This function seems redundant. If we are JWT is the source of truth. If we are checking that, don't think we need this function.

internal/auth/auth.go
```diff
@@ -344,3 +352,68 @@ func persistConfiguration() error {
 	_ = os.MkdirAll(filepath.Dir(path), os.ModePerm)
 	return os.WriteFile(path, data, 0o600)
 }
+
+func RefreshCloudSession() error {
+	if globalConfig == nil {
```
abhisek: I see we are duplicating this too many times. May be just add this in `init(...)`, after `_ = loadConfiguration(..)` and remove from everywhere else. So that we can guarantee that `globalConfig` will never be `nil`

internal/auth/auth.go
```diff
@@ -344,3 +352,68 @@ func persistConfiguration() error {
...
+		c := DefaultConfig()
+		globalConfig = &c
+	}
+
```
abhisek, line 375: I think we should check if `globalConfig.CloudRefreshToken` is empty in which case just return an error that `cannot refresh session: require new login session`

abhisek (changes requested):

internal/auth/auth.go
Sahilb315: The `cloud_access_token_updated_at` property allows us to quickly check if we should verify the token, without parsing the JWT every time (which could slow down commands). That’s why we use a short threshold (5 min) and only parse the JWT occasionally. If we remove `ShouldCheckAccessTokenExpiry` and always parse the JWT, we could increase the threshold (e.g., to 30 min or more), since we’d be doing the slower, accurate check less frequently.

cmd/cloud/main.go
Sahilb315: https://github.com/safedep/vet/pull/565#discussion_r2276573592

cmd/cloud/main.go
```diff
@@ -45,3 +61,20 @@ func NewCloudCommand() *cobra.Command {
...
+
+func requireAccessTokenCheck(cmd *cobra.Command, args []string) error {
+	// Check if token was obtained/refreshed 5 mins ago
+	// If > 5 mins, check the access token expiry
+	// else return
+	if auth.ShouldCheckAccessTokenExpiry() {
+		// Check if access token is expired
+		// If expired (ok), refresh the session
+		if ok, err := auth.IsAccessTokenExpired(); err != nil {
+			return fmt.Errorf("failed to check access token expiry: %w", err)
```
abhisek, line 79: @Sahilb315 I tested the case where the existing refresh token is invalid. As expected, the JWT parsing will rightly fail. But the error message is confusing. It does not tell the user to re-login.

I think since `requireAccessTokenCheck` is used in our command layer which in turn interact with the user, we should print appropriate error using `ui.PrintError` function in addition to the existing behaviour of printing the error.

I suggest we print an error: `Automatic token refresh failed, please re-login using vet cloud login --tenant <X>`

The value of `X` can be `TenantDomain()` if available to make it easy for users, else a placeholder like `<your-tenant-domain>`

Image

Sahilb315: @abhisek Made the changes.

PR #567, opened 2025-08-18
add `IsSuspicious` value for reporting suspicious packages too

### LINKED ISSUE #562: Sync Reporter does not report Malicious Packages correctly
We can verify this by scanning `vet` `go.mod` itself using `vet`

Following command required `vet` to be configured with API key

```
vet scan -M go.mod --insights-v2 --report-sync --report-sync-project safedep/vet --report-sync-project-version demo/product-flow-discussion -l- -d | grep -i "malware: true"
```

Here you will see the sync reporter do not sync anything with `malware=true` even though the console reporter flags multiple suspicious packages.

Note: We consider all `malware=true` as suspicious by default unless `verified=true` as well, in which case we consider the package to be malicious.

Sahilb315:
Fixes #562


abhisek:
Simple fix for supposedly complicated problems :) 

Good job @Sahilb315

PR #571, opened 2025-08-21
add support for generating sbom for homebrew installed packages

### LINKED ISSUE #568: Support SBOM generation for installed Homebrew packages
Add support for generating SBOMs from installed Homebrew packages, e.g.: `brew info --installed --json` to output an SBOM format for auditing and compliance.

Sahilb315:
Fixes #568

### DIFF
--- a/pkg/models/models.go
+++ b/pkg/models/models.go
@@ -44,6 +44,7 @@ const (
 	ManifestSourceLocal         = ManifestSourceType("local")
 	ManifestSourcePurl          = ManifestSourceType("purl")
 	ManifestSourceGitRepository = ManifestSourceType("git_repository")
+	ManifestSourceHomebrew      = ManifestSourceType("homebrew")
 )
 
 // We now have different sources from where a package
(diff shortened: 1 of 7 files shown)


pkg/models/models.go
```diff
@@ -141,6 +142,14 @@ func NewPackageManifestFromGitHub(repo, repoRelativePath, realPath, ecosystem st
 	}, realPath, ecosystem)
 }
 
+func NewPackageManifestFromHomebrew() *PackageManifest {
+	return newPackageManifest(PackageManifestSource{
+		Type:      ManifestSourceHomebrew,
+		Namespace: "brew.sh",
+		Path:      "homebrew",
+	}, "brew info --installed --json", "homebrew")
```
abhisek: Let us please define a constant for the ecosystem here.
https://github.com/safedep/vet/blob/main/pkg/models/models.go#L38

We also need an equivalent ecosystem in our API spec
https://github.com/safedep/api

So that we can map vet ecosystem to Control Tower ecosystem:
https://github.com/safedep/vet/blob/main/pkg/models/models.go#L218

abhisek (changes requested):

scan.go
```diff
@@ -231,6 +232,7 @@ func newScanCommand() *cobra.Command {
 	cmd.Flags().BoolVarP(&scanImageNoRemote, "image-no-remote", "", false,
 		"Disable container image pulling when not found locally")
 	cmd.Flags().StringVar(&reportHtmlPath, "report-html", "", "Path to write HTML report output")
+	cmd.Flags().BoolVar(&brewSpec, "brew", false, "Enable scanning for Homebrew packages")
```
abhisek: Lets please make the opt name `homebrew` because it becomes a contract and we can't change it later. Lets be explicit to avoid ambiguity with something else named `brew`

abhisek:
@Sahilb315 Can you please rebase this branch with main?

scan.go
Sahilb315: Makes sense. Renamed it

abhisek (approved):

PR #587, opened 2025-09-02
fix: mcp SSE hardening

arunanshub:


mcp/server/guard.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"net/http"
+	"strings"
+)
+
+// hostGuard is a middleware that allows only the allowed hosts to access
+// the MCP server.
+func hostGuard(next http.Handler) http.Handler {
+	allowedHosts := map[string]struct{}{
```
abhisek: These values must be user configurable with our opinionated defaults.

mcp/server/guard.go
```diff
@@ -0,0 +1,46 @@
...
+			w.WriteHeader(http.StatusMisdirectedRequest)
+			return
+		}
+		next.ServeHTTP(w, r)
+	})
+}
+
+// originGuard is a middleware that allows only the allowed origins to access
+// the MCP server.
+func originGuard(next http.Handler) http.Handler {
```
abhisek: Both the guards need an integration test. See `see_integration_test.go`

mcp/server/sse.go
```diff
@@ -38,6 +38,9 @@ func NewMcpServerWithSseTransport(config McpServerConfig) (*mcpServer, error) {
 
 			// Wrap the SSE server with HEAD request support
 			wrappedHandler := sseHandlerWithHeadSupport(s)
+			wrappedHandler = originGuard(wrappedHandler)
```
abhisek: Need docs update. See `docs/mcp.md`

mcp/server/guard.go
```diff
@@ -0,0 +1,70 @@
...
+		"http://127.0.0.1:",
+		"https://localhost:",
+	}
+	defaultAllowedHosts = []string{"localhost:9988", "127.0.0.1:9988", "[::1]:9988"}
+)
+
+// hostGuard is a middleware that allows only the allowed hosts to access the
+// MCP server. nil allowedHosts will use the default allowed hosts.  Empty
+// allowedHosts will block all hosts.
+func hostGuard(allowedHosts []string, next http.Handler) http.Handler {
```
abhisek: It should take `allowedHosts` from config

mcp/server/guard.go
```diff
@@ -0,0 +1,70 @@
...
+	if allowedHosts == nil {
+		allowedHosts = defaultAllowedHosts
+	}
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// contains is faster than a map lookup for small lists
+		if !slices.Contains(allowedHosts, r.Host) {
+			// 421 (misdirected request) is ideal; 403 (forbidden) is fine too.
+			w.WriteHeader(http.StatusMisdirectedRequest)
```
abhisek: Lets go with `forbidden` please. Remove the unnecessary comment

mcp/server/guard.go
```diff
@@ -0,0 +1,70 @@
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		o := r.Header.Get("Origin")
+		if o == "" {
+			// Non-browser/same-origin fetches may omit Origin. Don't block
+			// solely on this.
+			next.ServeHTTP(w, r)
+			return
+		}
+
+		if allowedOriginsPrefix == nil {
```
abhisek: Use from config

mcp/server/guard.go
```diff
@@ -0,0 +1,70 @@
...
+			http.Error(w, "forbidden origin", http.StatusForbidden)
+			return
+		}
+
+		next.ServeHTTP(w, r)
+	})
+}
+
+// isAllowedOrigin checks if the origin is in the allowed origins prefix list.
+func isAllowedOrigin(origin string, allowedOriginsPrefix []string) bool {
```
abhisek, line 49: Is this separate function required? We can just inline this because its not reusable.

mcp/server/sse_integration_test.go
```diff
@@ -17,13 +18,26 @@ func TestSSEServerIntegration(t *testing.T) {
 	mcpServer := server.NewMCPServer("test-vet-mcp", "0.0.1",
 		server.WithInstructions("Test MCP server for integration testing"))
 
-	// Create SSE server with our custom handler
+	// Create SSE server with our custom handler and wrap with guards
 	sseServer := server.NewSSEServer(mcpServer, server.WithStaticBasePath(""))
```
abhisek, line 22: Test with config

mcp/server/sse_integration_test.go
```diff
@@ -47,6 +63,37 @@ func TestSSEServerIntegration(t *testing.T) {
 		assert.True(t, resp.ContentLength <= 0, "HEAD request should not have content length > 0")
 	})
 
+	t.Run("GET request with invalid host should be blocked", func(t *testing.T) {
```
abhisek, line 71: Test for cases where config have allowed origin servers

abhisek (changes requested):

mcp/server/guard.go
arunanshub: I mean it looks cleaner as a separate function (following clean code). Plus go compiler already inlines it:

```text
# github.com/safedep/vet/mcp/server
/home/linuxbrew/.linuxbrew/Cellar/go/1.25.0/libexec/src/slices/slices.go:96:6: can inline slices.Index[go.shape.[]string,go.shape.string]
/home/linuxbrew/.linuxbrew/Cellar/go/1.25.0/libexec/src/slices/slices.go:117:6: can inline slices.Contains[go.shape.[]string,go.shape.string]
mcp/server/guard.go:21:6: can inline hostGuard
mcp/server/guard.go:27:26: can inline hostGuard.func1
mcp/server/guard.go:65:6: can inline isAllowedOrigin
...
```

Can inline this function manually if you want.

mcp/server/guard.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	defaultAllowedHosts = []string{"localhost:9988", "127.0.0.1:9988", "[::1]:9988"}
+)
+
+// hostGuard is a middleware that allows only the allowed hosts to access the
+// MCP server. nil allowedHosts will use the default allowed hosts.  Empty
+// allowedHosts will block all hosts.
+func hostGuard(config McpServerConfig, next http.Handler) http.Handler {
+	allowedHosts := config.SseServerAllowedHosts
+	if config.SseServerAllowedHosts == nil {
```
abhisek: This can be empty as well right? `len(..) == 0` is probably better

mcp/server/guard.go
```diff
@@ -0,0 +1,72 @@
...
+		next.ServeHTTP(w, r)
+	})
+}
+
+// originGuard is a middleware that allows only the allowed origins to access
+// the MCP server. nil allowedOriginsPrefix will use the default allowed origins
+// prefix. Empty allowedOriginsPrefix will block all origins.
+func originGuard(config McpServerConfig, next http.Handler) http.Handler {
+	allowedOriginsPrefix := config.SseServerAllowedOriginsPrefix
+	if config.SseServerAllowedOriginsPrefix == nil {
```
abhisek: Check for empty array

mcp/server/sse.go
```diff
@@ -34,10 +34,16 @@ func NewMcpServerWithSseTransport(config McpServerConfig) (*mcpServer, error) {
...
 		servingFunc: func(srv *mcpServer) error {
 			logger.Infof("Starting MCP server with SSE transport: %s", config.SseServerAddr)
-			s := server.NewSSEServer(srv.server, server.WithStaticBasePath(config.SseServerBasePath))
+			s := server.NewSSEServer(
+				srv.server,
+				server.WithStaticBasePath(config.SseServerBasePath),
+			)
 
 			// Wrap the SSE server with HEAD request support
 			wrappedHandler := sseHandlerWithHeadSupport(s)
```
abhisek, line 51: `Access-Control-Allow-Origin: *` in `sseHandlerWithHeadSupport`. It needs to be handled by checking and validating origin and only allowing the requested origin after verification.

mcp/server/guard.go
arunanshub: We distinguish between `nil` and empty array. Only when the array is `nil` is when we apply the default values. We consider empty array as "block all hosts/origins". Wdyt?

mcp/server/guard.go
arunanshub: same here: We distinguish between `nil` and empty array. Only when the array is `nil` is when we apply the default values. Wdyt?

mcp/server/sse.go
```diff
@@ -14,15 +14,23 @@ func sseHandlerWithHeadSupport(handler http.Handler) http.Handler {
...
 			w.Header().Set("Content-Type", "text/event-stream")
 			w.Header().Set("Cache-Control", "no-cache")
 			w.Header().Set("Connection", "keep-alive")
-			w.Header().Set("Access-Control-Allow-Origin", "*")
+
+			// Set CORS headers based on the request origin (will be validated
+			// by originGuard middleware)
+			if origin := r.Header.Get("Origin"); origin != "" {
+				w.Header().Set("Access-Control-Allow-Origin", origin)
+			}
```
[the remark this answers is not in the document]
arunanshub: it does occur first. In fact, we validate host -> origin -> set header.

mcp/server/sse.go
arunanshub: Done. Our middlewares go like host -> origin -> ACAO Header

mcp/server/server.go
```diff
@@ -24,6 +28,18 @@ func DefaultMcpServerConfig() McpServerConfig {
 		// SSE server will automatically add `/sse` to the base path
 		SseServerBasePath: "",
 		SseServerAddr:     "localhost:9988",
+
+		// By default, we use the current hardcoded values for backwards compatibility
+		// Users can customize these lists as needed for their deployment environment
+		SseServerAllowedHosts: []string{"localhost:9988", "127.0.0.1:9988", "[::1]:9988"},
```
abhisek, line 34: Whats the point of these defaults if you have separate defaults in `guard.go`? These defaults are never used. I think we should use these defaults and not have separate definition in `guard.go`.

Override these defaults only when user supplied values (slice) is not empty.

abhisek (changes requested):

mcp/server/server.go
arunanshub: Ah makes sense. Will do that.

PR #603, opened 2025-09-22
skips analytics init for root or help cmd/flags

### LINKED ISSUE #601: The help message should be printed without extra delay
Currently there is a small delay when we do `vet --help`. This is caused due to posthog reporting flush.

Sahilb315:
Fixes #601

### DIFF
--- a/main.go
+++ b/main.go
@@ -66,26 +66,16 @@ func main() {
 		Short:            "[ Establish trust in open source software supply chain ]",
 		TraverseChildren: true,
 		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
-			// Skip analytics for help and root-without-subcommand
-			if cmd.Flags().Changed("help") {
-				return nil
-			}
-			if cmd.Name() == "help" {
-				return nil
-			}
-
-			// Check any -h/--help on any subcommand by scanning argv
-			for _, a := range os.Args[1:] {
-				if a == "-h" || a == "--help" {
-					return nil
-				}
-			}
+			// Skip analytics for bare `vet` (no subcommand selected)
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -65,6 +65,29 @@ func main() {
 		Use:              "vet [OPTIONS] COMMAND [ARG...]",
 		Short:            "[ Establish trust in open source software supply chain ]",
 		TraverseChildren: true,
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			// Skip analytics for help and root-without-subcommand
+			if cmd.Flags().Changed("help") {
```
abhisek: What does this condition check?

main.go
```diff
@@ -65,6 +65,29 @@ func main() {
...
+				return nil
+			}
+			if cmd.Name() == "help" {
+				return nil
+			}
+
+			// Check any -h/--help on any subcommand by scanning argv
+			for _, a := range os.Args[1:] {
```
abhisek: This looks too hackish. 

Also looking at `analytics.go` it seems to me that `init()` will auto-initialize the analytics which will cause the delay. We have to use a custom `Init()` or something like which is called explicitly. So that analytics events are ignore unless explicitly initialized

abhisek: I would rather prefer a clean implementation. If that means, manually initializing `analytics.Init()` and `analytics.traceCommandRun, analytics.TraceCI` per command. But its good to avoid duplication.

Sahilb315: @abhisek Fixed it. `PersistentPreRunE` runs only on non-help commands & on root command (eg. `vet`). For bare metal root cmd we do a check & for other cases it isnt executed which is the help commands case

PR #613, opened 2025-10-05
feat: show manifest relative path in summary report

### LINKED ISSUE #503: Show source file for vulnerability findings.
We have data for `location` (source file location) for vulnerability. We need it to find which file found the package.

KunalSin9h:
fixes #503 

image

### DIFF
--- a/pkg/reporter/summary.go
+++ b/pkg/reporter/summary.go
@@ -572,7 +572,7 @@ func (r *summaryReporter) addRemediationAdviceTableRows(tbl table.Writer,
 
 	currentWorkingDirectory, err := os.Getwd()
 	if err != nil {
-		log.Warnf(WarningText(fmt.Sprintf("Error getting current working directory: %s", err.Error())))
+		log.Warnf("Error getting current working directory: %s", err.Error())
 		currentWorkingDirectory = ""
 	}
 
@@ -645,11 +645,10 @@ func (r *summaryReporter) addRemediationAdviceTableRows(tbl table.Writer,
 			}
 		}
 
-		manifestPath, err := r.packageManifestRelativePath(sp.pkg, currentWorkingDirectory)
-		// If failed to get relative path, fallback to absolute path
+		manifestPath, err := r.packageManifestRelativePath(currentWorkingDirectory, sp.pkg.Manifest.Path)
 		if err != nil {
-			log.Warnf(WarningText(fmt.Sprintf("Error getting manifest relative path: %s", err.Error())))
-			manifestPath = sp.pkg.Manifest.Path
+			log.Warnf("Error getting manifest relative path: %s", err.Error())
+			manifestPath = sp.pkg.Manifest.GetDisplayPath()
 		}
 
 		// Add Manifest Path information just below package name
@@ -691,10 +690,8 @@ func (r *summaryReporter) packageVulnerabilityRiskText(pkg *models.Package) stri
 	return WhiteBgText(" Unknown ")
 }
 
-func (r *summaryReporter) packageManifestRelativePath(pkg *models.Package, currentWorkingDirectory string) (string, error) {
-	fullPath := pkg.Manifest.Path
-
-	relPath, err := filepath.Rel(currentWorkingDirectory, fullPath)
+func (r *summaryReporter) packageManifestRelativePath(currentWorkingDirectory, manifestFullPath string) (string, error) {
+	relPath, err := filepath.Rel(currentWorkingDirectory, manifestFullPath)
 	if err != nil {
 		return "", fmt.Errorf("error getting relative path: %w", err)
 	}
--- a/pkg/reporter/summary_test.go
+++ b/pkg/reporter/summary_test.go
@@ -3,7 +3,6 @@ package reporter
 import (
 	"testing"
 
-	"github.com/safedep/vet/pkg/models"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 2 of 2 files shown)


pkg/reporter/summary.go
```diff
@@ -667,6 +689,17 @@ func (r *summaryReporter) packageVulnerabilityRiskText(pkg *models.Package) stri
 	return text.BgWhite.Sprint(" Unknown ")
 }
 
+func (r *summaryReporter) packageManifestRelativePath(pkg *models.Package, currentWorkingDirectory string) (string, error) {
```
abhisek: What will be the experience if I am scanning say /tmp/x? Also what about other readers like pURL, GitHub and all supported readers?

KunalSin9h: Scanning from outside the root, we will show relative path from current working directory. 

image

KunalSin9h: > pURL, GitHub and all supported readers?

readers? why any concern to readers?

KunalSin9h: for purl reader, we will default to `pkg.Manifest.Path` 

image

pkg/reporter/summary.go
```diff
@@ -637,6 +644,22 @@ func (r *summaryReporter) addRemediationAdviceTableRows(tbl table.Writer,
 			}
 		}
 
+		manifestPath, err := r.packageManifestRelativePath(sp.pkg, currentWorkingDirectory)
+		// If failed to get relative path, fallback to absolute path
+		if err != nil {
+			fmt.Println(text.FgRed.Sprint("error getting manifest relative path: ", err.Error()))
```
abhisek: Remove this error msg. May be just log it as warning

pkg/reporter/summary.go
```diff
@@ -568,6 +569,12 @@ func (r *summaryReporter) addRemediationAdviceTableRows(tbl table.Writer,
 		return tagText
 	}
 
+	currentWorkingDirectory, err := os.Getwd()
+	if err != nil {
+		fmt.Println(text.FgRed.Sprint("Error getting current working directory: ", err.Error()))
```
abhisek: log it as warning

pkg/reporter/summary.go
```diff
@@ -637,6 +644,22 @@ func (r *summaryReporter) addRemediationAdviceTableRows(tbl table.Writer,
...
+		// If failed to get relative path, fallback to absolute path
+		if err != nil {
+			fmt.Println(text.FgRed.Sprint("error getting manifest relative path: ", err.Error()))
+			manifestPath = sp.pkg.Manifest.Path
+		}
+
+		// Add Manifest Path information just below package name
+		tbl.AppendRow(table.Row{
+			"", // Ecosystem
+			text.FgBlue.Sprint(manifestPath),
```
abhisek: Use the utils from the color profile instead of hardcoding the color.

KunalSin9h:
@abhisek resolved comments, now using standard color profile. and logging error as warning.

pkg/reporter/summary.go
```diff
@@ -637,6 +645,22 @@ func (r *summaryReporter) addRemediationAdviceTableRows(tbl table.Writer,
 			}
 		}
 
+		manifestPath, err := r.packageManifestRelativePath(sp.pkg, currentWorkingDirectory)
+		// If failed to get relative path, fallback to absolute path
+		if err != nil {
+			log.Warnf(WarningText(fmt.Sprintf("Error getting manifest relative path: %s", err.Error())))
```
abhisek: WarningText / formatting should not be used for logging

pkg/reporter/summary.go
```diff
@@ -568,6 +570,12 @@ func (r *summaryReporter) addRemediationAdviceTableRows(tbl table.Writer,
 		return tagText
 	}
 
+	currentWorkingDirectory, err := os.Getwd()
+	if err != nil {
+		log.Warnf(WarningText(fmt.Sprintf("Error getting current working directory: %s", err.Error())))
```
abhisek: WarningText / formatting should not be used for logging

pkg/reporter/summary.go
```diff
@@ -667,6 +691,17 @@ func (r *summaryReporter) packageVulnerabilityRiskText(pkg *models.Package) stri
 	return WhiteBgText(" Unknown ")
 }
 
+func (r *summaryReporter) packageManifestRelativePath(pkg *models.Package, currentWorkingDirectory string) (string, error) {
+	fullPath := pkg.Manifest.Path
```
abhisek: Check that it is a valid local path using `os.Stat` and if not, return error early?

abhisek (changes requested):

pkg/reporter/summary.go
```diff
@@ -637,6 +645,22 @@ func (r *summaryReporter) addRemediationAdviceTableRows(tbl table.Writer,
 			}
 		}
 
+		manifestPath, err := r.packageManifestRelativePath(sp.pkg, currentWorkingDirectory)
+		// If failed to get relative path, fallback to absolute path
+		if err != nil {
+			log.Warnf(WarningText(fmt.Sprintf("Error getting manifest relative path: %s", err.Error())))
+			manifestPath = sp.pkg.Manifest.Path
```
abhisek: I think we should fall back to `Manifest.GetDisplayPath()`

Thats because it is readers responsibility to set the display path correctly. Since reporter is user facing, it should use `DisplayPath` instead of the `Path` which can be internal.

Image

abhisek (changes requested):

pkg/reporter/summary.go
KunalSin9h: done

pkg/reporter/summary.go
KunalSin9h: No need for its, as if its not a valid path then `filepath.Rel` will fail since it can't find relative path, then we fall back to `pkg.Manifest.GetDisplayPath()`.

abhisek (approved):

PR #615, opened 2025-10-06
fix: recommended way os using golangci lint

KunalSin9h:


CONTRIBUTING.md
```diff
@@ -34,25 +34,31 @@ When contributing changes to repository, follow these steps:
 
 ### Requirements
 
-* Go 1.24.3+
+- Go 1.24.3+
```
abhisek, line 37: Update this to `1.25+`

abhisek: Accepting. Lets do this in next PR.

PR #617, opened 2025-10-08
feat: hidden command for manual generation

KunalSin9h:
https://www.notion.so/safedep-inc/SafeDep-Vet-manual-pages-28661d70b236806cbd81ef2b12d95ee9


cmd/manual/main.go
```diff
@@ -0,0 +1,71 @@
...
+
+var (
+	// mdManualOutDir is the output directory for markdown manual files
+	mdManualOutDir string
+
+	// manManualOutDir is the output directory for troff (man markup) manual files
+	manManualOutDir string
+)
+
+func NewManualCommand() *cobra.Command {
```
abhisek: `Manual` word is ambiguous here. Which means `done manually or with hand`.

I suggest an experience like this:

```
./vet doc generate --man ./dir1 --markdown ./dir2
```

At least one of the output directory should be required. Both should not be enforced. Even if its an internal command, I think we should consider proper user experience.

So may be create a `doc` command and `generate` as a sub-command of it.

PR #621, opened 2025-10-10
fix: github pages CI build jekyll and deploy to new service

KunalSin9h:


.github/workflows/gh-pages-deploy.yml
```diff
@@ -32,8 +44,33 @@ jobs:
       - name: Run diff check against ${{ env.SOURCE_GEN_DIR }}
         run: bash ./scripts/diff_docgen.sh ${{ env.SOURCE_GEN_DIR }} ${{ env.TEMP_GEN_DIR }}
 
-      - name: Deploy to Github Pages
-        uses: peaceiris/actions-gh-pages@v4
+  # Build jenkll (md -> html)
+  build:
```
abhisek, line 48: This should run after checking drift. If drift check fails, we should not run build

.github/workflows/gh-pages-deploy.yml
```diff
@@ -5,13 +5,25 @@ on:
     branches:
       - main
 
+# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
+permissions:
```
abhisek, line 9: Also need CI check for drift detection and failure. So that dev knows they need to run docgen

KunalSin9h: that is already happening in `check-generated-drift` job

KunalSin9h: see previous pr failed on this: https://github.com/safedep/vet/actions/runs/18407952478/job/52452342842

abhisek: I mean when PR is raised. This job runs on push right?

KunalSin9h: oh, got it.

KunalSin9h: will add after things will work.

PR #625, opened 2025-10-13
fix: Handle Multiple License Type in npm Graph Parser

### LINKED ISSUE #191: npm Graph Parser is Broken for Lockfile Version v3
`vet` fails to parse `package-lock.json` v3 with error:

```
scan failed due to error: json: cannot unmarshal array into Go struct field npmPackageLockPackage.packages.license of type string
```

This seems to be because in v3, `license` is an array instead of a string value.

Ref: https://docs.npmjs.com/cli/v9/configuring-npm/package-lock-json

Sahilb315:
Fixes #191

### DIFF
--- a/pkg/parser/npm_graph.go
+++ b/pkg/parser/npm_graph.go
@@ -12,22 +12,60 @@ import (
 	"github.com/safedep/vet/pkg/models"
 )
 
-// jsonString is a lenient string type that ignores non-string JSON values when unmarshaling.
-// This helps with fields like "license" that can appear as objects/arrays from older versions which are deprecated.
-type jsonString string
+// https://docs.npmjs.com/cli/v10/configuring-npm/package-json#license
+type npmLicenseObject struct {
+	Type string `json:"type"`
+	URL  string `json:"url"`
+}
 
-func (s *jsonString) UnmarshalJSON(b []byte) error {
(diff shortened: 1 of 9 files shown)


pkg/parser/npm_graph.go
```diff
@@ -12,9 +12,22 @@ import (
 	"github.com/safedep/vet/pkg/models"
 )
 
+// jsonString is a lenient string type that ignores non-string JSON values when unmarshaling.
+// This helps with fields like "license" that can appear as objects/arrays from older versions which are deprecated.
+type jsonString string
+
+func (s *jsonString) UnmarshalJSON(b []byte) error {
```
abhisek: Since we are doing this, might as well handle the case of `[{"type": "...", "url":""}]` as a fallback ie. when string unmarshalling fails, we can try to unmarshall as array of object and take only the 1st one from that entry and use the `.type` from it.

Also we should name it as `npmLicenseType` because we are using this specifically for handling the the license type inconsistency in `package-lock.json` parsing.

Should have independent test cases for this custom type to avoid subtle bugs.

abhisek:
@KunalSin9h I made some enhancements in this branch. Can you please review?

Sahilb315:
@abhisek LGTM. You will need to approve once & we can merge

PR #642, opened 2025-11-26
fix: verify mcp publish job failing

KunalSin9h:
The publishing mcp job passes and mcp data is [published](https://registry.modelcontextprotocol.io/v0.1/servers?search=io.github.safedep/vet-mcp). [Action Here](https://stackoverflow.com/questions/68808854/environment-variables-not-always-being-expanded-in-github-actions-workflow-file).

But The verification job failed due to a common GitHub Action issue of not **expanding the other envs in `env:`** blok [issue reference](https://stackoverflow.com/questions/68808854/environment-variables-not-always-being-expanded-in-github-actions-workflow-file). 

Also it fixes one version issue which is caused by registry UI, since the official registry has prefix of `v` in its UI, and since we are publishing with version as `v***`, it looks like `vv**`. Hence we use version as `1.x.x` and for docker image we have hardcoded `v` in its identifier. 

image

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,7 +34,7 @@ When contributing changes to repository, follow these steps:
 
 ### Requirements
 
-- Go 1.24.3+
+- Go 1.25.0+
 
 ### Install Dependencies
 
--- a/lefthook.yml
+++ b/lefthook.yml
@@ -7,6 +7,8 @@ pre-push:
 pre-commit:
   parallel: true
   commands:
+    linter-fmt:
+      run: golangci-lint fmt
     linter:
       run: golangci-lint run -n
     secrets-scanning:


.github/workflows/container.yml
```diff
@@ -115,14 +115,15 @@ jobs:
...
-        run: echo "VET_VERSION=${GITHUB_REF#refs/tags/}" >> $GITHUB_ENV
+        run: echo "VET_VERSION=${GITHUB_REF#refs/tags/v}" >> $GITHUB_ENV
 
       - name: Query MCP Registry and verify server is published
         env:
           SERVER_NAME: "io.github.safedep/vet-mcp"
-          EXPECTED_VERSION: $VET_VERSION
           REGISTRY_URL: "https://registry.modelcontextprotocol.io/v0.1/servers"
         run: |
+          export EXPECTED_VERSION=$VET_VERSION
```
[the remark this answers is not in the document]
KunalSin9h: it will work

.github/workflows/container.yml
```diff
@@ -88,7 +88,7 @@ jobs:
         run: sudo apt-get update && sudo apt-get install -y jq
 
       - name: Get version from tag
```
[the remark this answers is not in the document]
KunalSin9h: done

.github/workflows/container.yml
```diff
@@ -88,7 +88,10 @@ jobs:
...
-        run: echo "VET_VERSION=${GITHUB_REF#refs/tags/}" >> $GITHUB_ENV
+        # Strip 'v' prefix from tag (e.g., v1.0.0 -> 1.0.0) as 
+        # - we want clean version (x.y.z) without v prefix, since its already added by registry UI
+        # - in case of docker image, we hardcod in server.json docker image identifier
```
abhisek:
```suggestion
        # - in case of docker image, we hardcode in server.json docker image identifier
```

.github/workflows/container.yml
```diff
@@ -115,14 +118,18 @@ jobs:
...
```
abhisek:
```suggestion
        # - in case of docker image, we hardcode in server.json docker image identifier
```

.mcp-publisher/server.json
```diff
@@ -32,7 +32,7 @@
   "packages": [
     {
       "registryType": "oci",
-      "identifier": "ghcr.io/safedep/vet:VERSION_FROM_ENV",
+      "identifier": "ghcr.io/safedep/vet:vVERSION_FROM_ENV",
```
abhisek, line 35: Why add `v` here? I see the entry published in the registry is `vv1.2.3`

KunalSin9h: previously our value for `VERSION_FROM_ENV` was `v1.x.y`, for both mcp version and identifier version. But we are now using `VERSION_FROM_ENV` as `1.x.y` so that our mcp version is clean (without `v`) and hence when the registry add its `v` prefex then it become normal `v1.x.y` and not `vv.1.x.y`. BUT since our `VERSION_FROM_ENV` is without `v`. and our docker image need `v` i have hardcoded. it.

KunalSin9h: @abhisek `vv1.2.3` is due to top level `.version` not this OCI identifier.

abhisek (approved):

PR #647, opened 2025-12-03
Fix CycloneDX SBOM validation

### LINKED ISSUE #637: SBOM Report Produces non-compliant SBOM
## Background

User reported issue:

> Using the sbom generated by vet. Using this to validate the schema https://cyclonedx.github.io/cyclonedx-web-tool/validate

## Requirement

`vet` should produce SBOM that is valid and verifiable using official SBOM schema validator.

Sahilb315:
Fixes #637

### DIFF
--- a/pkg/reporter/cyclonedx.go
+++ b/pkg/reporter/cyclonedx.go
@@ -45,8 +45,7 @@ type cycloneDXReporter struct {
 	rootComponentBomref       string
 	bomEcosystems             map[string]bool
 	bomVulnerabilitiesBomrefs map[string]bool
-
-	bomPackageRef map[string]struct{}
+	bomPackageRef             map[string]bool
 }
 
 var cdxUUIDRegexp = regex.MustCompileAndCache(`^urn:uuid:[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$`)
@@ -111,7 +110,7 @@ func NewCycloneDXReporter(config CycloneDXReporterConfig) (Reporter, error) {
 		rootComponentBomref:       rootComponentBomref,
 		bomEcosystems:             map[string]bool{},
 		bomVulnerabilitiesBomrefs: map[string]bool{},
-		bomPackageRef:             map[string]struct{}{},
+		bomPackageRef:             map[string]bool{},
 	}, nil
 }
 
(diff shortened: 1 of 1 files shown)


pkg/reporter/cyclonedx.go
```diff
@@ -110,6 +111,7 @@ func NewCycloneDXReporter(config CycloneDXReporterConfig) (Reporter, error) {
 		rootComponentBomref:       rootComponentBomref,
 		bomEcosystems:             map[string]bool{},
 		bomVulnerabilitiesBomrefs: map[string]bool{},
+		bomPackageRef:             map[string]struct{}{},
```
abhisek:
```suggestion
		bomPackageRef:             map[string]bool{},
```

Lets just do bool like the others please.

pkg/reporter/cyclonedx.go
```diff
@@ -188,15 +199,9 @@ func (r *cycloneDXReporter) resolvePackageLicenses(pkg *models.Package) []cdx.Li
 	for _, license := range *pkg.Insights.Licenses {
 		licenseChoice := cdx.LicenseChoice{
 			License: &cdx.License{
-				Name: string(license),
```
abhisek, line 191: What is the issue with name? Does the spec says not to have name?

pkg/reporter/cyclonedx.go
```diff
@@ -188,15 +199,9 @@ func (r *cycloneDXReporter) resolvePackageLicenses(pkg *models.Package) []cdx.Li
...
-				ID:   string(license),
+				ID: string(license),
 			},
 		}
-		spdxLicense, availableSpdxLicense := data.SpdxLicenses[string(license)]
-		if availableSpdxLicense {
```
abhisek, line 196: We should have these license metadata if available right? What does the CycloneDX spec says?

pkg/reporter/cyclonedx_test.go
```diff
@@ -285,8 +285,6 @@ func TestCycloneDxReporterLicenses(t *testing.T) {
 	licenses := utils.SafelyGetValue(utils.SafelyGetValue(generatedBom.Components)[0].Licenses)
 	assert.Len(t, licenses, 1)
 	assert.Equal(t, "MIT", licenses[0].License.ID)
-	assert.Equal(t, "MIT License", licenses[0].License.Name)
```
abhisek, line 288: Not sure if we should just remove them.

pkg/reporter/cyclonedx.go
Sahilb315: Yeah, Only one of `ID` or `Name` can be present. No other values can be present.

Schema to verify: https://cyclonedx.org/schema/bom-1.6.schema.json
You can search for `"license":` to find the licenses schema

pkg/reporter/cyclonedx.go
Sahilb315: Only one of `ID` or `Name` can be present. No other values can be present.

Schema to verify: https://cyclonedx.org/schema/bom-1.6.schema.json
You can search for `"license":` to find the licenses schema

abhisek (approved):

PR #649, opened 2025-12-09
Only scan private repos under org based scan

infosecwonderland:
### Title:
add filters for GitHub org scan (private only, exclude forks & archived)

### Summary
This PR adds more control to vet scan `--github-org` by allowing users to:

- Exclude forked repositories

- Skip archived repositories

- Scan only private repositories in an organization

`--github-org-include-archived (default: true)`
Include archived repositories when scanning a GitHub organization.

`--github-org-exclude-forks (default: false)`
Exclude forked repositories from the org scan.

`--github-org-private-only (default: false)`
Only scan private repositories in the organization.

### Motivation

When running `org-wide` scans in larger GitHub organizations, archived repositories and forks can introduce noise and increase scan time. Many security teams (including ours) primarily care about active, private, non-fork repos.

### DIFF
--- a/pkg/readers/github_org_reader.go
+++ b/pkg/readers/github_org_reader.go
@@ -23,8 +23,8 @@ type GithubOrgReaderConfig struct {
 	MaxRepositories        int
 	SkipDependencyGraphAPI bool
 	ExcludeRepos           []string
-	PrivateOnly            bool // only scan private repos
-	ExcludeForks           bool // skip forked repos    
+	PrivateOnly            bool 
+	ExcludeForks           bool   
 }
 
 type githubOrgReader struct {


pkg/readers/github_org_reader.go
```diff
@@ -23,6 +23,8 @@ type GithubOrgReaderConfig struct {
 	MaxRepositories        int
 	SkipDependencyGraphAPI bool
 	ExcludeRepos           []string
+	PrivateOnly            bool
```
abhisek, line 26: Looks like formatting issue.

scan.go
```diff
@@ -150,6 +153,12 @@ func newScanCommand() *cobra.Command {
 		"Maximum number of repositories to process for the Github Org")
 	cmd.Flags().StringArrayVarP(&githubOrgExcludedRepos, "github-org-exclude-repos", "", []string{},
 		"Comma-separated list of GitHub repos to exclude during org scan (format: org/repo1,org/repo2)")
+	cmd.Flags().BoolVarP(&githubOrgIncludeArchived, "github-org-include-archived", "", true,
```
abhisek, line 156: I think we should flip this ie. `github-org-exclude-archive` (default: false) so that it aligns with `github-org-exclude-forks` naming convention.

abhisek: Never mind. I think `github-org-include-archived` makes sense. But we should set the default value to `false`. The rationale is, as an opinion, we lean towards not including archived repositories. However, a single flag can be used to enable it.

abhisek:
Fixed PR: https://github.com/safedep/vet/pull/650

PR #661, opened 2026-01-23
feat: added malware info filed in json report

KunalSin9h:


api/json_report_spec.proto
```diff
@@ -92,11 +92,18 @@ message PackageReport {
...
   repeated InsightLicenseInfo licenses = 6;
   repeated InsightProjectInfo projects = 8;
+  repeated MalwareInfo malware_info = 9;
 
   // Threats
   repeated ReportThreat threats = 7;
 }
 
+message MalwareInfo {
+  string type = 1;
```
abhisek: Type should be an enum right?

api/json_report_spec.proto
```diff
@@ -92,11 +92,18 @@ message PackageReport {
...
+  string confidence = 2;
```
abhisek: Same for confidence ie. Enum

api/json_report_spec.proto
```diff
@@ -92,11 +92,32 @@ message PackageReport {
...
 
+enum MalwareType {
+  MALWARE_TYPE_UNSPECIFIED = 0;
+  SAFE = 1;
+  SUSPICIOUS = 2;
+  MALICIOUS = 3;
+}
+
+enum MalwareConfidence {
+  MALWARE_CONFIDENCE_UNSPECIFIED = 0;
```
abhisek: The naming convention is inconsistent. Any reason for this?

KunalSin9h: @abhisek yes it not consistent, fixed it.

PR #671, opened 2026-02-02
update `vet inspect malware` to exit with proper error

Sahilb315:
With verbose mode:
image

Without verbose mode:
image

### DIFF
--- a/cmd/inspect/malware.go
+++ b/cmd/inspect/malware.go
@@ -48,7 +48,7 @@ func newPackageMalwareInspectCommand() *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			err := executeMalwareAnalysis()
 			if err != nil {
-				ui.PrintError("Failed: %v", err)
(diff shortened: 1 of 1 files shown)


KunalSin9h (dismissed):

cmd/inspect/malware.go
abhisek: @Sahilb315 I think we should replace this with `tui.ErrorExit`. At the top level. What do you think?

cmd/inspect/malware.go
```diff
@@ -148,7 +149,7 @@ func executeMalwareAnalysis() error {
 		},
 	})
 	if err != nil {
-		return fmt.Errorf("failed to submit package for malware analysis: %v", err)
```
abhisek, line 151: I think we should return the error here because this is the function that handles the flow orchestration. Let it return its own error but change `%v` to `%w` so that it is wrapped.

Let ErrorExit handled at the top level so that it can handle other possible error cases as well that we have standardised in DRY

PR #674, opened 2026-02-03
fix: Handle 429 Error in Enrich Malware in vet and Report Quota Error after Scan Completion

KunalSin9h:
fixes https://github.com/safedep/control-tower/issues/758

image

Screenshot_2026-02-03_11-56-40

---

<a href="https://app.devin.ai/review/safedep/vet/pull/674">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/pkg/reporter/markdown_summary.go
+++ b/pkg/reporter/markdown_summary.go
@@ -599,9 +599,3 @@ func (m *markdownSummaryMalwareInfo) renderMalwareInfoTable() (string, error) {
 
 	return tbl.RenderMarkdown(), nil
 }
-
-func renderMarkdownQuotaLimitErrorMessages(quotaExceededErrCnt int) string {
-	return fmt.Sprintf("⚠️ You have reached your **quota** for on-demand malicious package "+
-		"scanning. `%d` on-demand analysis requests were **denied**. Please see [safedep.io/pricing](https://safedep.io/pricing) for "+
-		"upgrade.", quotaExceededErrCnt)
-}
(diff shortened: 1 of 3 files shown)


pkg/reporter/markdown_summary.go
```diff
@@ -589,3 +599,9 @@ func (m *markdownSummaryMalwareInfo) renderMalwareInfoTable() (string, error) {
 
 	return tbl.RenderMarkdown(), nil
 }
+
+func renderMarkdownQuotaLimitErrorMessages(quotaExceededErrCnt int) string {
```
abhisek: Should this be moved to `common.go` so that the same can be used in other markdown reports as well if required? Since this function is naked ie. not within the struct, I am guessing this is re-usable.

pkg/scanner/enrich_malware.go
```diff
@@ -262,6 +269,20 @@ func (e *malysisMalwareEnricher) submitPackageForAnalysis(pkg *models.Package) (
 
 	res, err := e.client.AnalyzePackage(ctx, &req)
 	if err != nil {
+		st, ok := status.FromError(err)
```
abhisek: Why not use `usefulerror.AsUsefulError(..)` instead of manually checking for code. 

I see we have support for this.

https://github.com/safedep/dry/blob/main/usefulerror/convert_grpc.go#L94

abhisek (changes requested):

KunalSin9h:
@abhisek resolved comments.

PR #675, opened 2026-02-03
Feat/759 auth swith query when not entitled

KunalSin9h:
fixes: https://github.com/safedep/control-tower/issues/759

flows:
image

result:

image

image

image

---

<a href="https://app.devin.ai/review/safedep/vet/pull/675">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/pkg/reporter/summary_test.go
+++ b/pkg/reporter/summary_test.go
@@ -64,7 +64,7 @@ func TestRenderInternalErrorMessages(t *testing.T) {
 		r := &summaryReporter{}
 		r.internalReportConfig.malwareAnalysisQuotaLimitErrorCount = 5
 
-		expectedErrorMessage := "You have reached your quota for on-demand malicious package scanning. 5 on-demand analysis requests were denied. Please see safedep.io/pricing for upgrade."
+		expectedErrorMessage := "You have reached your quota for on-demand malicious package scanning.\n5 on-demand analysis requests were denied. Please see safedep.io/pricing for upgrade."
 
 		actualErrorMessage := renderQuotaLimitErrorMessages(5)
 


internal/auth/entitlements.go
```diff
@@ -0,0 +1,99 @@
...
+	"fmt"
+	"sync"
+
+	"buf.build/gen/go/safedep/api/grpc/go/safedep/services/controltower/v1/controltowerv1grpc"
+	v11 "buf.build/gen/go/safedep/api/protocolbuffers/go/safedep/messages/controltower/v1"
+	controltowerv1 "buf.build/gen/go/safedep/api/protocolbuffers/go/safedep/services/controltower/v1"
+	"github.com/safedep/vet/pkg/common/logger"
+)
+
+type Entitlement struct {
```
abhisek: This is already defined in protobuf generate code. Why are we duplicating this?

internal/auth/entitlements.go
```diff
@@ -0,0 +1,99 @@
...
+	cachedEntitlements := make([]Entitlement, len(response.GetEntitlements()))
+	for i, entitlement := range response.GetEntitlements() {
+		cachedEntitlements[i] = Entitlement{
+			Feature: entitlement.Entitlement.Feature,
+			Limit:   entitlement.Entitlement.Limit,
+		}
+	}
+
+	// Cache the entitlements
+	globalEntitlementsManager.mu.Lock()
```
abhisek: This is mixing concerns. We should call `globalEntitlementsManager.load(entitlements)` which should internally lock and set its values.

internal/auth/entitlements.go
```diff
@@ -0,0 +1,99 @@
...
+	globalEntitlementsManager.mu.RLock()
+	defer globalEntitlementsManager.mu.RUnlock()
+
+	if !globalEntitlementsManager.loaded {
+		logger.Debugf("Entitlements not loaded, returning false")
+		return false
+	}
+
+	// Check if all requested entitlements are available
+	for _, requestedEntitlementFeature := range entitlementsFeatures {
```
abhisek: This should be inside `globalEntitlementsManager`. We should do something like `globalEntitlementsManager.hasEntitlement(...)`

pkg/reporter/common.go
```diff
@@ -73,3 +74,8 @@ func renderMarkdownQuotaLimitErrorMessages(quotaExceededErrCnt int) string {
 		"scanning. `%d` on-demand analysis requests were **denied**. Please see [safedep.io/pricing](https://safedep.io/pricing) for "+
 		"upgrade.", quotaExceededErrCnt)
 }
+
+func renderMarkdownEntitlementAutoSwitchEnabled() string {
+	return fmt.Sprintln("🔀 You are **not entitled** for **on-demand** malicious package scanning." +
+		" The scan is auto-configured to use malicious package **query** only. See [safedep.io/pricing](https://safedep.io/pricing) for upgrade.")
```
abhisek: Re-thinking on the copy text, I think `entitled` is a SaaS specific terminology. Not the most user friendly.

Can you update the copy to something like this:

```
On-demand malicious package scanning is not available on the Free plan. Your scan was configured to use  known malicious packages feed.  Upgrade: https://safedep.io/pricing
```

cmd/inspect/malware.go
```diff
@@ -82,12 +83,19 @@ func newPackageMalwareInspectCommand() *cobra.Command {
...
+	err := auth.LoadEntitlements()
 	if err != nil {
 		return fmt.Errorf("access to Malicious Package Analysis requires an API key. " +
 			"For more details: https://docs.safedep.io/cloud/quickstart/")
 	}
 
+	// Check if tenant has malware analysis entitlement
+	found := auth.HasEntitlements(controltowerv1.Feature_FEATURE_ACTIVE_MALICIOUS_PACKAGE_SCANNING)
+	if !found {
+		return fmt.Errorf("your current plan does not include Active Malicious Package Analysis. " +
```
abhisek:
```suggestion
		return fmt.Errorf("your current plan does not include on-demand malicious package analysis. " +
```

internal/auth/entitlements.go
```diff
@@ -0,0 +1,100 @@
...
+	mu           sync.RWMutex
+	loaded       bool
+	entitlements []v1.Entitlement
+}
+
+// Global instance of the entitlements manager
+var globalEntitlementsManager = &entitlementsManager{}
+
+// cache caches the entitlements, ie save them
+func (g *entitlementsManager) cache(entitlements []v1.Entitlement) {
```
abhisek:
```suggestion
func (g *entitlementsManager) store(entitlements []v1.Entitlement) {
```

Cache is not the right intention here. Its more of `store`.

internal/auth/entitlements.go
```diff
@@ -0,0 +1,100 @@
...
+	g.mu.Lock()
+	defer g.mu.Unlock()
+
+	g.loaded = true
+	g.entitlements = entitlements
```
abhisek: What if we call `store` multiple times, can we add to the list while avoiding duplicate?

Note: Should have a test case for this scenario

scan.go
```diff
@@ -923,39 +914,52 @@ func internalStartScan() error {
 				"For more details: https://docs.safedep.io/cloud/quickstart/")
 		}
 
-		client, err := auth.MalwareAnalysisClientConnection("vet-malware-analysis")
-		if err != nil {
-			return err
-		}
+		found := auth.HasEntitlements(controltowerv1.Feature_FEATURE_ACTIVE_MALICIOUS_PACKAGE_SCANNING)
+		if !found {
+			ui.PrintWarning("You are not entitled for on-demand malicious package scanning. The scan is auto-configured to use malicious package query only. See safedep.io/pricing for upgrade.")
```
abhisek: Use similar copy text as suggested earlier

scan.go
```diff
@@ -923,39 +914,52 @@ func internalStartScan() error {
...
+		} else {
+			malwareEnricher, err := createMalwareAnalysisEnricher(githubClient)
+			if err != nil {
+				return fmt.Errorf("failed to create malware enricher: %w", err)
+			}
 
-		malwareEnricher, err := scanner.NewMalysisMalwareEnricher(client, githubClient, config)
-		if err != nil {
-			return err
+			ui.PrintMsg("Using Malysis for malware analysis")
```
abhisek: Fix the copy text to indicate if it is using `on-demand` malware analysis. Also remove `Malysis`, which is an internal service.

abhisek (changes requested):

scan.go
```diff
@@ -923,39 +914,52 @@ func internalStartScan() error {
...
-			return err
-		}
+		found := auth.HasEntitlements(controltowerv1.Feature_FEATURE_ACTIVE_MALICIOUS_PACKAGE_SCANNING)
+		if !found {
+			ui.PrintWarning("%s", getOnDemandMalwareAnalysisMissingEntitlementMessage())
 
-		config := scanner.DefaultMalysisMalwareEnricherConfig()
-		config.Timeout = malwareAnalysisTimeout
+			// auto-switch to enrichMalwareQuery mode
+			switchToEnrichQueryMalware = true
```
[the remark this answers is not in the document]
KunalSin9h: i think is is expected, when user specify `--no-verify-auth` then we are /should consider no auth (no entitlement), hence when --malware is given we should switch to query mode. 

@abhisek wdyt?

cmd/inspect/malware.go
```diff
@@ -82,12 +83,19 @@ func newPackageMalwareInspectCommand() *cobra.Command {
...
 
-	err := auth.Verify()
+	err := auth.LoadEntitlements()
 	if err != nil {
 		return fmt.Errorf("access to Malicious Package Analysis requires an API key. " +
 			"For more details: https://docs.safedep.io/cloud/quickstart/")
 	}
 
+	// Check if tenant has malware analysis entitlement
+	found := auth.HasEntitlements(controltowerv1.Feature_FEATURE_ACTIVE_MALICIOUS_PACKAGE_SCANNING)
```
abhisek: Is this check required here? I believe we already have `ErrorExit` based on error from API call.

What do you think @Sahilb315 ?

pkg/reporter/markdown_summary.go
```diff
@@ -101,6 +102,9 @@ func NewMarkdownSummaryReporter(config MarkdownSummaryReporterConfig) (Reporter,
 		malwareInfo: &markdownSummaryMalwareInfo{
 			malwareInfo: make(map[string]*markdownSummaryPackageMalwareInfo),
 		},
+		internalErrorCounter: internalErrorCounter{
```
abhisek: `internalErrorCounter` is no longer the correct name now. Why not just name it as `internalReportConfig` if you need this shared across reporters?

pkg/reporter/common.go
```diff
@@ -73,3 +74,8 @@ func renderMarkdownQuotaLimitErrorMessages(quotaExceededErrCnt int) string {
 		"scanning. `%d` on-demand analysis requests were **denied**. Please see [safedep.io/pricing](https://safedep.io/pricing) for "+
 		"upgrade.", quotaExceededErrCnt)
 }
+
+func renderMarkdownEntitlementAutoSwitchEnabled() string {
```
abhisek, line 78: We should add this for the console reporter as well? The one that is rendered by default when we run `vet scan ..`

scan.go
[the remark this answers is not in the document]
abhisek: @KunalSin9h Yes. Agreed.

pkg/reporter/common.go
KunalSin9h: no, since in console / summary, we are printing the auto-switch message when we actually auto-switching at `scan.go` itself., i.e no in reporters, before any reporting. its for markdown summary only. 

That `ui.print` in `scan.go` prints in terminal for before summary / console, hence we dont need any duplicate message in these reporters.

cmd/inspect/malware.go
KunalSin9h: yes, no need for it.

Sahilb315:
All above mentioned cases successfully pass.

scan.go
```diff
@@ -346,7 +348,10 @@ func startScan() {
 
 func internalStartScan() error {
 	if agentSkillSpec != "" {
-		return runAgentSkillScan()
+		if err := runAgentSkillScan(); err != nil {
+			tui.ErrorExit(err)
+			return err
+		}
 	}
```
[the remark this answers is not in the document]
KunalSin9h: fixed.

scan.go
```diff
@@ -1060,34 +1067,20 @@ func runAgentSkillScan() error {
...
-
-		enricher, err = scanner.NewMalysisMalwareEnricher(client, githubClient, config)
+		malwareEnricher, err := createMalwareAnalysisEnricher(githubClient)
 		if err != nil {
 			return fmt.Errorf("failed to create malware enricher: %w", err)
 		}
 
+		enricher = malwareEnricher
 		scanMode = "active"
 	}
```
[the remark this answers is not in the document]
KunalSin9h: that's intentional!,

pkg/reporter/common.go
```diff
@@ -58,12 +58,13 @@ func getVulnerabilitySolution(pkg *models.Package) string {
...
+type internalReportConfig struct {
+	malwareAnalysisQuotaLimitErrorCount         int
+	malwareAnalysisEntitlementAutoSwitchEnabled bool
 }
 
 func renderQuotaLimitErrorMessages(quotaExceededErrCnt int) string {
-	return fmt.Sprintf("You have reached your quota for on-demand malicious package "+
+	return fmt.Sprintf("You have reached your quota for on-demand malicious package.\n"+
 		"scanning. %d on-demand analysis requests were denied. Please see safedep.io/pricing for "+
 		"upgrade.", quotaExceededErrCnt)
```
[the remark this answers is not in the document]
KunalSin9h: fixed.

PR #719, opened 2026-05-07
feat: vet endpoint scan

arunanshub:
---

<a href="https://app.devin.ai/review/safedep/vet/pull/719" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


cmd/endpoint/scan.go
```diff
@@ -0,0 +1,333 @@
...
+	silent       bool
+	kinds        []string
+	drainTimeout time.Duration
+}
+
+// AliasOptions is the public surface for cmd packages that re-export
+// `vet endpoint scan` under a friendlier name. It excludes --kind on
+// purpose: an alias pins the scanner kind via its dedicated entrypoint
+// (e.g. RunAITool), it does not pick the kind set itself.
+type AliasOptions struct {
```
abhisek: `scanOptions` and `AliasOptions` seems to be exact duplicate. Whats the rationale for violating DRY here? What do we gain?

cmd/endpoint/scan.go
```diff
@@ -0,0 +1,333 @@
...
+func buildCloudSink(_ context.Context, creds auth.Credentials, opts scanOptions) (inventory.Sink, func(), error) {
+	conn, err := dialSyncConn(creds)
+	if err != nil {
+		return nil, nil, fmt.Errorf("dial sync endpoint: %w", err)
+	}
+
+	transport := endpointsync.NewGrpcTransport(conn)
+	identity := endpointsync.NewEndpointIdentityResolver()
+
+	client, err := endpointsync.NewSyncClient(
```
abhisek, line 254: The `NewSyncClient` provides a `WithWALPath(..)` option, without which it falls back to a default path. I am thinking if we should explicitly set the WAL path to a tmp path and explicitly delete the WAL after the scan finishes as part of a closer function.

My rationale is, `vet endpoint scan` is really a scanner. It collects information and syncs it. Unlikely PMG, where every event have a security sensitivity, `vet endpoint scan` does not. Given this, does it makes sense to re-sync events from a previously failed scan?

Current behaviour guarantees that all events are synced. What happens when `vet endpoint scan` is run without cloud credentials? It will likely keep growing the WAL unnecessarily ?

docs/adr/0001-endpoint-scan.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Context
+
+vet has a local AI tool discoverer (`vet ai discover`) whose output never reaches SafeDep Cloud. We need a per-endpoint inventory in the Endpoint Hub UI. Cloud-side ingestion (`VetInventoryEvent`, `vet_inventory_events` table, denormalised filter columns) and the Hub list shipped to prod on 2026-05-07. This ADR fixes the vet-side surface: command shape, sync model, credentials, and the producer architecture that subsequent endpoint scanners will share.
+
+## Decision
+
+### Command
+
+- New top-level: `vet endpoint scan`. Scans the endpoint, syncs to cloud when credentials are present.
```
abhisek, line 14: What happens when credentials are not present?

docs/adr/0001-endpoint-scan.md
```diff
@@ -0,0 +1,85 @@
...
+- `vet ai discover` is kept as a permanent alias for `vet endpoint scan --kind ai-tool`.
+- No `--no-sync` flag. Behaviour is determined by credential presence:
```
abhisek, line 16: What if credentials are present but they are incorrect. What is the behaviour?

docs/adr/0001-endpoint-scan.md
```diff
@@ -0,0 +1,85 @@
...
+  - credentials present: scan, stream events to the WAL, background flusher delivers to cloud, command exits after discovery completes and the WAL is drained (bounded grace).
+  - credentials absent: scan, render the local table, exit 0 with a one-line hint pointing at `safedep auth login` or `SAFEDEP_API_KEY`.
```
abhisek: So all open source users, who runs `vet endpoint scan` will end up having a bloated local database for no reason?

docs/adr/0001-endpoint-scan.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Scope of `vet endpoint scan`
+
+In scope, by milestone:
+
+- M1: AI tools (existing `pkg/aitool` discoverers).
+- Near term: IDE extensions (already emitted by `pkg/aitool`), browser extensions.
+- Medium term: OS packages (brew, dnf, apt, choco), full filesystem OSS package inventory.
```
abhisek: Add file system level OSS package discovery as a future milestone. We need a way to scan entire filesystem to detect `npm`, `pypi` and other support packages in local file system.

Actually doing it is not hard, but doing it with efficiency and without causing performance friction for the user working on the endpoint is challenging.

docs/adr/0001-endpoint-scan.md
```diff
@@ -0,0 +1,85 @@
...
+- Each discovered item produces one `ITEM_OBSERVED` event. End of scan emits one `SCAN_SUMMARY` event. Errors during scan emit `ERROR` events.
+- Two goroutines, one durable buffer:
+  - The producer goroutine (the command) runs scanners and writes events to the local `endpointsync` WAL (a SQLite file). Each write is a local insert, on the order of microseconds. The producer never blocks on the network.
+  - The consumer goroutine lives inside `endpointsync.SyncClient`. It drains the WAL and ships batches to cloud over gRPC. It runs concurrently with the producer.
+  - The WAL is the boundary. It survives process crashes; events queued before a crash ship on the next run.
+- End of scan: the command calls `Close` on the cloud sink, which waits for the WAL to drain up to a bounded deadline (default 30s, configurable). If the deadline hits, undelivered events stay in the WAL for the next run and the command exits 0.
+
+### Credentials
+
+1. vet env vars (`VET_API_KEY`, `SAFEDEP_API_KEY`, `VET_INSIGHTS_API_KEY`, plus tenant equivalents).
```
abhisek: What about `SAFEDEP_TENANT_ID` ?

docs/adr/0001-endpoint-scan.md
```diff
@@ -0,0 +1,85 @@
...
+2. vet's `~/.safedep/vet-auth.yml` (existing legacy file; respected if present but won't be extended).
+3. DRY keychain provider, constructed without an insecure file fallback.
+4. DRY env provider.
+5. Fail with a hint: "no SafeDep credentials; run `safedep auth login` or set `SAFEDEP_API_KEY`."
```
abhisek: `safedep auth login` ? `vet` should not know anything about CLI. It is an independent OSS tool and should continue being independent.

docs/adr/0001-endpoint-scan.md
```diff
@@ -0,0 +1,85 @@
...
+
+vet's `vet-auth.yml` is the legacy insecure store. It is not deprecated by this ADR but no new write paths are added to it. Headless and WSL environments are served by env vars.
+
+### Producer architecture
+
+A new package, `pkg/inventory`, owns the producer pipeline. Three concepts:
+
+- `Item`: domain struct mirroring the proto `VetInventoryEvent.ItemObserved` 1:1 (kind, identity, source id, name, app, scope, config path, enabled, optional typed details, metadata). Scanners emit `*Item`. The wire proto is not used as the in-process type.
+- `Scanner`: `Scan(ctx, ScanConfig, emit func(*Item) error) error`. The orchestrator iterates a `[]Scanner`. Adding a new scanner kind is one new implementation plus one registration.
+- `Sink`: `Begin(Session) → Emit(*Item) → End(ScanSummary) → Close`. The orchestrator owns session lifecycle (invocation_id, started_at, counts), fans each item to all sinks, builds the summary. M1 sinks: `LocalSink` (table, `--report-json`), `CloudSink` (wraps `endpointsync.SyncClient`).
```
abhisek, line 65: We will soon have a need to produce a CycloneDX SBOM based on Endpoint Scan.

docs/adr/0001-endpoint-scan.md
```diff
@@ -0,0 +1,85 @@
...
+
+  Parallel scanner execution is not in M1; if a future scanner's runtime justifies it, parallelism slots in via `errgroup` over `[]Scanner` and sinks gain a thread-safety requirement at that point.
+
+- `CloudSink.Emit` writes to the WAL and returns. It does not call gRPC, does not retry, does not block on the network. Retry and backoff are owned by `endpointsync`.
+
+- Transient sink errors (a WAL write that fails, a cloud delivery rejection inside the flusher) are logged and do not abort the scan. Only fatal or programming errors propagate up.
+
+- The orchestrator never sees `*AITool` or wire proto types. Translation `*AITool` to `*Item` lives in one adapter (`pkg/inventory/scanners/aitool`); translation `*Item` to wire proto lives in `CloudSink`.
+
+For M1, AI tool discovery is wired in by a single `Scanner` that wraps `aitool.Registry.Discover`. Existing readers under `pkg/aitool` are unchanged.
```
abhisek, line 79: How will it expand to IDE Extension discovery?

cmd/endpoint/scan.go
```diff
@@ -0,0 +1,333 @@
...
+
+	sinks := []inventory.Sink{deps.buildLocalSink(opts)}
+
+	cloudCleanup := func() {}
+	creds, resolveErr := deps.resolver.Resolve(ctx)
+	switch {
+	case resolveErr == nil:
+		cloudSink, cleanup, err := deps.buildCloudSink(ctx, creds, opts)
+		if err != nil {
+			_, _ = fmt.Fprintf(stderr, "vet endpoint scan: failed to enable cloud sync (%v); continuing with local-only output\n", err)
```
abhisek: Writing directly to terminal breaks our path for DRY/TUI convergence in future. Use `internal/ui` utils from `vet` itself for now. We will replace them with our common TUI in future.

abhisek: Why is this an error. What if a user wants to run this only for local purpose?

abhisek (changes requested):

cmd/endpoint/scan.go
arunanshub: Local continues to work.

docs/adr/0001-endpoint-scan.md
arunanshub: ahh oops I meant to write `vet cloud login`. Must have been thinking about safedep-cli when I was writing this.

docs/adr/0001-endpoint-scan.md
arunanshub: Note that on a successful sync, the WAL is emptied out successfully anyway.

docs/adr/0001-endpoint-scan.md
arunanshub: We'll just implement the scanner interface. It's the scanner that decides where to look for and how to make sense of IDE extension. Yes it's a bit complex since ide extensions are compressed, but nevertheless follow a standard: https://learn.microsoft.com/en-us/visualstudio/extensibility/anatomy-of-a-vsix-package?view=visualstudio

cmd/endpoint/scan.go
arunanshub: Makes sense, I wanted to use dry, but half-dry half-internal felt like a wrong idea since it'll leave the codebase in a half-migrated state. Using internal for now.

docs/adr/0001-endpoint-scan.md
arunanshub: Scan will happen, WAL will NOT be written to.

cmd/endpoint/scan.go
arunanshub: > What happens when vet endpoint scan is run without cloud credentials? It will likely keep growing the WAL unnecessarily ?

Nothing actually. In fact my plan was to at least allow the data to be accumulated in the WAL, plus by default the WAL is strictly bounded. Then when the user does auth, this audit trail that has been in a pending state can be shipped along with any new data.

Plus if you think about it, having a temporary WAL defeats the point of an audit trail. Consider this (assume user is auth'd):

- t=0: user's device is offline. scan happens. vscode detected.
- t=10: user uninstalls vscode.
- t=20: user is back online. scan happens. vscode not detected.

If we have had temporary WAL, we would have never known that between t=0 and t=20, vscode was uninstalled. Having a persistent WAL keeps the value of the audit trail. Wdyt @abhisek?

cmd/endpoint/scan.go
arunanshub: fixed

cmd/endpoint/scan.go
abhisek: We agreed that we will keep the WAL only for authenticated users so that WAL is never created for OSS users.

abhisek (approved):

PR #722, opened 2026-05-08
fix: agent skills discovery and syncing

KunalSin9h:
Screenshot_2026-05-08_18-45-30

Screenshot_2026-05-08_18-44-54

### DIFF
--- a/pkg/inventory/scanners/skills/adapter.go
+++ b/pkg/inventory/scanners/skills/adapter.go
@@ -4,75 +4,121 @@ import (
 	"context"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/safedep/vet/pkg/inventory"
 )
 
-const scannerName = "skills"
+const (
+	scannerName = "skills"
 
-// agentEntry maps an agent identifier to its project-local and global skill
-// directory paths. ProjectPath is relative to cfg.ProjectDir; GlobalPath is
-// relative to cfg.HomeDir.
-type agentEntry struct {
-	App         string
-	ProjectPath string
-	GlobalPath  string
+	// appGlobalSkills is the App name used for skill directories that are
+	// shared by multiple agents (e.g. .agents/skills/). Using a single name
+	// avoids emitting the same skill once per agent that claims the path.
+	appGlobalSkills = "Global Skills"
+)
+
+// pathEntry is a (relative path, app name) pair for one scope.
+// Path is relative to cfg.ProjectDir (project scope) or homeDir (system scope).
+type pathEntry struct {
+	App  string
+	Path string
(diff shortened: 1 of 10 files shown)


pkg/inventory/scanners/skills/adapter.go
```diff
@@ -0,0 +1,133 @@
...
+type agentEntry struct {
+	App         string
+	ProjectPath string
+	GlobalPath  string
+}
+
+// agentRegistry is the canonical mapping of agents to their skill directories.
+// Derived from the supported-agents table in the agent skills scanner design doc.
+var agentRegistry = []agentEntry{
+	{App: "amp", ProjectPath: ".agents/skills", GlobalPath: ".config/agents/skills"},
```
abhisek: `.agent/skills` is common for a LOT of agents. I think we should use a common `App` name for it. Goal for this scanner is to discover Skills.

pkg/inventory/scanners/skills/adapter.go
```diff
@@ -0,0 +1,133 @@
...
+func scanDir(_ context.Context, dir, app string, scope inventory.Scope, emit inventory.EmitFunc) error {
+	entries, err := os.ReadDir(dir)
+	if err != nil {
+		return nil
+	}
+	for _, e := range entries {
+		if !e.IsDir() {
+			continue
+		}
+		s := &skill{
```
abhisek, line 242: Is there any SKILL metadata that we should read and capture? Check the skill spec once.

KunalSin9h: @abhisek `display_name`, `description` and `path` is the only guaranteed thinks, else are custom / agent specific names, its not standard so anyone an put anything in the front matter of the .md.

pkg/inventory/sinks/cloud/translate.go
```diff
@@ -43,14 +43,27 @@ func scanErrorToVetEvent(e inventory.ScanError) *controltowerv1pb.VetInventoryEv
 	}.Build()
 }
 
+// maxSourceIDLen is the backend API limit for the source_id field,
+// enforced via API validation error observed in production.
+const maxSourceIDLen = 100
+
+// truncateSourceID caps source_id to the backend's 100-character limit.
+func truncateSourceID(s string) string {
```
abhisek: @KunalSin9h Why is this required? Have you come across cases where Source ID is more than 100? The backend is our own. If there are valid cases, we can increase the limit.

KunalSin9h: fixed, removed it, unnecessary function.

abhisek (approved):

PR #727, opened 2026-05-14
feat: vscode and antigravity support in endpoint scan

KunalSin9h:


arunanshub:
### Code review

Found 1 issue:

1. **`vscode.go` system-scope MCP config path is wrong** — `~/.vscode/` is VS Code's extension cache directory, not the user configuration directory. Real VS Code user config (including `mcp.json`) lives at `~/.config/Code/User/` on Linux, `~/Library/Application Support/Code/User/` on macOS, and `%APPDATA%\Code\User\` on Windows. The current code will silently find nothing for system-scope MCP discovery on every real VS Code installation. Project-scope (`.vscode/mcp.json` relative to `projectDir`) is correct. Compare with `cursor.go` which uses `~/.cursor/` — VS Code's equivalent would be the user data dir, not the extensions dir.

https://github.com/safedep/vet/blob/db40c9e3b1fe1c88696195b80acb404f8314efa9/pkg/aitool/vscode.go#L39-L45

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

arunanshub (approved):

PR #729, opened 2026-05-15
fix: aggregate lockfile poisoning warnings per package+URL

### LINKED ISSUE #231: Avoid Duplicate Reporting of Lockfile Poisoning in Markdown Summary Reporter
This issue is manifested when there are a lot of packages from private repositories. Although we have a config option to add private repository URL to allow list, it makes sense to consolidate the finding by `URL`

Reference: https://github.com/safedep/vet-action/issues/37#issuecomment-2275005354

KunalSin9h:
fixes #231 

## Summary

When a package URL fails both the untrusted-host check and the path-convention check, vet was emitting **two separate warnings** for the same package and URL. This was redundant and noisy.

### After Changes. 

Screenshot_2026-05-15_16-18-21
Screenshot_2026-05-15_16-16-58


KunalSin9h:
@abhisek review please, fixed it myself, as the same issue which is solved by external contributor is not optimal.

arunanshub (approved):

PR #731, opened 2026-05-20
fix: better error message when auth fails

arunanshub:


internal/auth/verify.go
```diff
@@ -17,8 +24,19 @@ func Verify() error {
...
 	if err != nil {
-		return err
+		return wrapAuthError(err)
 	}
 
 	return nil
 }
+
+func wrapAuthError(err error) error {
+	if s, ok := status.FromError(err); ok {
```
abhisek: Why not use tui.ErrorExit from DRY or the usefulerror package from DRY? It is aware if different error types. 

Also i think we need to have a generic tui.ErrorExit for all cloud / auth related commands

arunanshub: Yes good idea. I think sooner or later vet would need a dry/tui overhaul anyway.

abhisek (approved):

PR #733, opened 2026-05-21
feat: migrate commands to cleaner error message

arunanshub:


internal/command/run.go
```diff
@@ -0,0 +1,32 @@
...
+	"fmt"
+
+	tuierrors "github.com/safedep/dry/tui/errors"
+	"github.com/safedep/dry/usefulerror"
+	"github.com/spf13/cobra"
+)
+
+// PresentableError converts a UsefulError into a human-readable error message.
+// Non-UsefulErrors are returned unchanged.
+func PresentableError(err error) error {
```
abhisek: Is this required to be package public?

arunanshub: fixed

abhisek (approved):

PR #741, opened 2026-06-01
feat: port vet to nx based release automation

arunanshub:


Sahilb315:
# Review Findings

## Blocker

- [ ] **Restore post-publish install verification.** No workflow at this PR's head installs the published package (verified: no `test-installation` job, `npm install -g`, or `npm view` in any of the 11 workflow files). The deleted `publish-npm.yml` had this coverage, and the pmg port kept it in `goreleaser.yml`: a `test-installation` job with `needs: release`, matrix of ubuntu/macos/windows and Node 18/20/22/24, `fail-fast: false`, a wait loop for registry propagation, then `npm install -g` and a smoke test. Without it, nothing ever tests the published packages on macOS or Windows: the `.exe` bin shim, the universal binary, and `optionalDependencies` resolution on a real install are unverified. Please port the pmg `test-installation` job as is.

## Testability / verifiability fixes

- [ ] **`sync-binaries:run` cache restores stale binaries locally (reproduced).** The target is cached (via `nx.json` targetDefaults) with inputs of only the sync tool sources and `dist/artifacts.json`. goreleaser's `artifacts.json` is byte-identical across Go source changes (verified with a real `goreleaser build --snapshot` run twice: Binary entries carry no version, commit, or checksum). Reproduced with nx `23.0.0-beta.11` in a minimal workspace mirroring this task graph: change the source, the goreleaser-equivalent task re-runs, but the sync task hits `[local cache]` and restores the *previous* binaries into `packages/*/bin`, so `vet:verify` smoke-tests a stale binary. The first change after a cold cache can mask it; every later source change restores stale output. CI is unaffected (fresh runners, and the release targets set `cache: false`), but the local loop documented in CONTRIBUTING.md validates the wrong binary. Fix: set `"cache": false` on `sync-binaries:run` (the copy is cheap; the expensive goreleaser task stays cached), or switch the input to `dependentTasksOutputFiles`. pmg has the identical latent bug, so the fix is worth backporting.

- [ ] **Wire `typecheck` into the verify chain (type errors ship silently today).** `@safedep/vet:typecheck` is defined, but no workflow or target invokes it (verified at head). tsdown does not type-check: verified with tsdown `0.22.0` and this PR's exact config, a file containing `const x: number = "not a number"` builds successfully and the broken code lands in `dist/bin.cjs`, while `tsc -p tsconfig.json --noEmit` correctly fails on it. Add `typecheck` as a dependency of `verify` or a step in `release-preflight`. Same gap exists in pmg.

## Release coordination (tap repo, not this PR)

- [ ] **Formula-to-cask switch needs the tap updated in the same release window.** `safedep/homebrew-tap` still serves `vet.rb` as a formula at the tap root (v1.18.1). Once this PR ships, goreleaser writes `Casks/vet.rb` and stops updating the formula. If the formula stays in place, `brew install safedep/tap/vet` keeps resolving to the frozen v1.18.1 formula, `brew upgrade` users never move past v1.18.1, and the name becomes ambiguous. The fix is the same two-step change pmg made in safedep/homebrew-tap@f7addcca: delete the root `vet.rb` and add `"vet": "vet"` to `tap_migrations.json`. Per the goreleaser migration guide, existing formula users then move to the cask automatically on `brew upgrade`, and `brew install safedep/tap/vet` keeps working unchanged (no `--cask` flag needed). Linux stays supported: the generated cask includes `on_linux` x86_64/arm64 binaries (see `Casks/pmg.rb`), which Homebrew installs on Linux since 4.5 (April 2025).

abhisek (approved):

PR #752, opened 2026-07-21
feat: Deprecate active malware analysis in favour of SafeDep threat intelligence query

abhisek:


scan.go
```diff
@@ -135,7 +134,7 @@ func newScanCommand() *cobra.Command {
 	cmd.Flags().BoolVarP(&enrichUsingInsightsV2, "insights-v2", "", true,
 		"Enrich package metadata using Insights V2 API")
 	cmd.Flags().BoolVarP(&enrichMalware, "malware", "", false,
-		"Enrich package metadata with active malware analysis results")
+		"Enrich package metadata with known malicious packages data (DEPRECATED: alias for --malware-query)")
```
KunalSin9h: **Silent downgrade for entitled `--malware` users — please make the deprecation loud at runtime.**

Previously, a user with the `FEATURE_ACTIVE_MALICIOUS_PACKAGE_SCANNING` entitlement running `--malware` got on-demand analysis of packages *not yet* in the database (zero-day coverage), and every fallback path printed a warning. After this PR, the same flag silently becomes a known-malware lookup: same invocation, same green scan output, strictly weaker detection, no notice. For a security tool this change should be impossible to miss.

Suggest using cobra's built-in mechanism, which prints a warning on every use and hides the flag from help (making the `(DEPRECATED: ...)` help-text suffix redundant):

```go
_ = cmd.Flags().MarkDeprecated("malware",
    "active (on-demand) analysis has been retired; this flag now behaves as --malware-query")
```

Optionally also a `ui.PrintWarning` in `internalStartScan` when `cmd.Flags().Changed("malware")`, so the notice lands in CI logs too.

---
_Generated by [Claude Code](https://claude.ai/code)_

pkg/reporter/common.go
```diff
@@ -59,8 +59,7 @@ func getVulnerabilitySolution(pkg *models.Package) string {
 }
 
 type internalReportConfig struct {
-	malwareAnalysisQuotaLimitErrorCount         int
-	malwareAnalysisEntitlementAutoSwitchEnabled bool
+	malwareAnalysisQuotaLimitErrorCount int
```
KunalSin9h, line 62: **Quota-limit reporting is now dead code — while query-side rate limits go unreported.**

`models.PackageManifest.IncrementMalwareAnalysisQuotaError()` had exactly one caller: the deleted `enrich_malware.go` (it incremented on `ErrQuotaExceeded`). After this PR nothing increments it, so `malwareAnalysisQuotaLimitErrorCount` is always 0 and the "`%d` on-demand analysis requests were **denied**" sections in `summary.go` and `markdown_summary.go` can never render.

Meanwhile the query path — especially the unauthenticated community connection — is rate-limited, and a `ResourceExhausted`/quota error from `QueryPackageAnalysis` now surfaces only as a generic wrapped enricher error with no report-level signal.

Two consistent options:
1. Handle quota errors in `malysisMalwareAnalysisQueryEnricher.Enrich` and call `IncrementMalwareAnalysisQuotaError`, keeping this reporter plumbing alive (and updating the message text, which still says "on-demand analysis"); or
2. Remove the whole path: this field, `renderQuotaLimitErrorMessages`, `renderMarkdownQuotaLimitErrorMessages`, the render sites in both reporters, and the related assertions in `summary_test.go`.

Keeping it half-alive is the worst option — it reads as functional but isn't.

---
_Generated by [Claude Code](https://claude.ai/code)_

pkg/scanner/enrich_malware_query.go
```diff
@@ -18,6 +19,34 @@ import (
...
+
+	// Number of workers to poll the analysis results. Retained for backward
+	// compatibility; unused by the query enricher.
+	QueryWorkerCount int
+
+	// Maximum number of retries for querying the analysis results for a given
+	// analysis identifier. Retained for backward compatibility; unused by the
+	// query enricher.
+	MaxQueryRetries int
+}
```
KunalSin9h, line 42: **`Timeout` is also unused — and that makes `--malware-analysis-timeout` a silent no-op.**

`QueryWorkerCount` and `MaxQueryRetries` are honestly documented as "retained for backward compatibility; unused", but `Timeout` — equally unread by this enricher — is documented as if functional. The only timeout actually applied is `GrpcOperationTimeout` (10s default) per `Enrich` call.

The knock-on effect is in `scan.go` (`createMalwareQueryEnricher`): `config.Timeout = malwareAnalysisTimeout` still wires the `--malware-analysis-timeout` flag into this field, so a user setting that flag now changes nothing, with no warning.

Suggest one of:
- wire the flag into `GrpcOperationTimeout` instead, or
- `MarkDeprecated` the `--malware-analysis-timeout` flag and mark `Timeout` unused here too.

Since the "malware enricher" this config was shared with is gone, consider pruning the struct down to just `GrpcOperationTimeout` — if the exported fields are being kept for external importers, it'd be worth saying so explicitly in the comment.

---
_Generated by [Claude Code](https://claude.ai/code)_

scan.go
```diff
@@ -910,41 +909,12 @@ func internalStartScan() error {
...
 
-	// used as flag in --report-malware-summary reporter to notify user about enricher auto-switching
-	switchedToEnrichQueryMalware := false
-
-	if enrich && enrichMalware {
+	// Active (on-demand) malware analysis has been deprecated. Both --malware and
+	// --malware-query now query SafeDep's threat intelligence database for known
+	// malicious packages. --malware is retained as a backward compatible alias.
+	if enrich && (enrichMalware || enrichMalwareQuery) {
 		analytics.TrackCommandScanMalwareAnalysis()
```
KunalSin9h, line 920: **Analytics semantics change: this event now fires on every default scan.**

On `main`, `TrackCommandScanMalwareAnalysis()` only fired when `--malware` was explicitly passed. Both `enrich` and `--malware-query` default to `true`, so this condition is true for a bare `vet scan` — `eventScanMalwareAnalysis` will jump to ~100% of scans. Any dashboard tracking malware-analysis adoption gets a discontinuity, and the event loses its signal.

Consider a distinct event for query enrichment, or firing this one only when the user explicitly opted in (`cmd.Flags().Changed("malware") || cmd.Flags().Changed("malware-query")`).

---
_Generated by [Claude Code](https://claude.ai/code)_

README.md
```diff
@@ -162,21 +159,21 @@ Source repositories: **GitHub**, **GitLab**
...
-vet scan -D . --malware
-
 # Query known malicious packages (no API key needed)
 vet scan -D . --malware-query
 ```
 
+> [!NOTE]
+> The `--malware` flag is deprecated. Active (on-demand) scanning has been retired in favour of
+> querying SafeDep's threat intelligence database. `--malware` now behaves identically to
+> `--malware-query` and is retained for backward compatibility.
```
KunalSin9h, line 175: **Nit: "retired" slightly overstates it** — `vet inspect malware` still performs full on-demand analysis (`AnalyzePackage` in `cmd/inspect/malware.go`), and this README still documents it a few lines below. Suggest adding a sentence here like: "On-demand analysis of individual packages remains available via `vet inspect malware`." It's accurate and gives users losing `--malware`'s active mode a pointer to the replacement.

---
_Generated by [Claude Code](https://claude.ai/code)_

KunalSin9h (commented):
Deep-dive review of the active-malware-analysis deprecation. Direction is good — this removes ~600 lines of complexity, the exclusion-record handling and GitHub Actions commit-SHA resolution survived intact in the query enricher, dropping the community-mode hard error means `--malware` now works keyless, and the new reporter copy is more honest than the old text. Branch builds cleanly and `go vet` passes; the deleted tests only covered the deleted enricher.

That said, two findings are significant and worth addressing before merge (inline comments):

1. **Silent security downgrade**: entitled `--malware` users lose on-demand/zero-day analysis with no runtime notice — needs `MarkDeprecated` and/or `ui.PrintWarning` (scan.go).
2. **Dead quota reporting**: `IncrementMalwareAnalysisQuotaError` has no callers after deleting `enrich_malware.go`, so the quota-denied report sections can never render, while real rate-limit errors from the query API surface as generic errors (pkg/reporter/common.go).

Smaller items inline: `--malware-analysis-timeout` is now a silent no-op, `eventScanMalwareAnalysis` now fires on every default scan (dashboard discontinuity), and a README nit about `vet inspect malware` still offering on-demand analysis.

One non-anchorable nit: the doc comment on `NewMalysisMalwareAnalysisQueryEnricher` ("We are re-using the config from the malware enricher because this enricher is a subset of the malware enricher") is now stale — that enricher no longer exists and the config lives here.

Also: consider linking this PR to a tracking issue for the active-analysis deprecation, since it changes paid-feature behavior.

---
_Generated by [Claude Code](https://claude.ai/code)_

scan.go
KunalSin9h: Fixed in f34e6dd. Added `cmd.Flags().MarkDeprecated("malware", ...)`, so using `--malware` now prints `Flag --malware has been deprecated, active (on-demand) analysis has been retired; this flag now behaves as --malware-query` on every invocation, and the flag is hidden from `scan --help`. Verified against a built binary.

---
_Generated by [Claude Code](https://claude.ai/code)_

pkg/reporter/common.go
KunalSin9h: Fixed in f34e6dd (option 1). The query enricher's `Enrich` now detects `usefulerror.ErrQuotaExceeded` from `QueryPackageAnalysis`, calls `IncrementMalwareAnalysisQuotaError()`, and continues to the next package instead of failing the scan — so the reporter quota sections reflect real denials again rather than being dead code. Also reworded the quota messages (and the matching `summary_test.go` assertion) to drop the "on-demand" phrasing.

---
_Generated by [Claude Code](https://claude.ai/code)_

pkg/scanner/enrich_malware_query.go
KunalSin9h: Fixed in f34e6dd. Marked `--malware-analysis-timeout` deprecated (it no longer affects the single-shot query lookup, which uses the fixed `GrpcOperationTimeout`), and updated the `Timeout` field doc to say it's retained for backward compatibility but unused — consistent with `QueryWorkerCount`/`MaxQueryRetries`. Left the struct fields in place rather than pruning, to avoid breaking external importers.

---
_Generated by [Claude Code](https://claude.ai/code)_

scan.go
KunalSin9h: Not changing this, by design. Now that known-malware querying is the default enrichment path, `eventScanMalwareAnalysis` firing on essentially every scan is arguably the correct semantic — malware analysis genuinely runs on every default scan. Suppressing it to only explicit opt-in would require threading `*cobra.Command` (and `Flags().Changed`) down into `internalStartScan`, which doesn't currently receive it. If per-flag adoption tracking is still wanted, that's better done as a separate, dedicated analytics event with input from whoever owns the dashboards — filing that as a follow-up rather than reworking the call chain here. Resolving.

---
_Generated by [Claude Code](https://claude.ai/code)_

README.md
KunalSin9h: Obsolete — won't apply. This suggestion was to point users at `vet inspect malware` as the on-demand replacement, but that command has since been deprecated too (commits 8c61894 / a6bc5f2), so directing users there would be wrong. The README now simply states active/on-demand scanning is retired and points to `--malware-query`, which is the correct guidance. Resolving.

---
_Generated by [Claude Code](https://claude.ai/code)_

KunalSin9h (dismissed):

cmd/inspect/malware.go
```diff
@@ -43,7 +43,13 @@ func newPackageMalwareInspectCommand() *cobra.Command {
...
-		Long:  `Inspect an OSS package for malware using SafeDep Malware Analysis API`,
+		Long: `Inspect an OSS package for malware using SafeDep Malware Analysis API.
+
+DEPRECATED: On-demand malware analysis is no longer available through this
+command. This command will be removed in a future release.`,
+		Deprecated: "on-demand malware analysis is no longer available through this command; " +
+			"it will be removed in a future release. Use `vet scan --malware-query` to check " +
+			"packages against SafeDep's known malicious packages database.",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			err := executeMalwareAnalysis()
```
abhisek: You need to remove the logic from `executeMalwareAnalysis()` and return a suitable useful error

KunalSin9h: Done in 3a98231. `executeMalwareAnalysis()` no longer performs any analysis — it now returns a `usefulerror` (`on_demand_malware_analysis_unavailable`) with a human message and help pointing to `vet scan --malware-query`. Removed the now-dead helpers (`writeOSVReport`, `writeJSONReport`, `renderMalwareAnalysisReport`, `reportVisualizationUrl`, `resolveGitHubActionsCommitHash`) and their imports, and routed `RunE` through `command.RunCmd` so the useful error renders properly (net −252/+12 lines). Runtime output:

```
[ERR] On-demand malware analysis is no longer available through `vet inspect malware`.: Use `vet scan --malware-query` to check packages against SafeDep's known malicious packages database.
```

---
_Generated by [Claude Code](https://claude.ai/code)_

KunalSin9h (approved):

vignesh2731 (approved):

PR #754, opened 2026-07-27
fix: preserve every resolved version when deduplicating packages

vignesh2731:
Key the deduplication on name and version instead. The existing behaviour is otherwise unchanged: entries with an unknown or empty version are still dropped. Rebuilding the slice in parse order also makes the output deterministic rather than dependent on Go map iteration order.

Note that the upstream pnpm parser is not at fault here. osv-scanner returns both is-number@6.0.0 and is-number@7.0.0 for the reproduction lockfile; vet discarded one of them.
Screenshot 2026-07-28 012001


pkg/readers/lockfile_reader.go
```diff
@@ -72,37 +72,7 @@ func (p *lockfileReader) EnumManifests(handler func(*models.PackageManifest,
...
-
-		// Convert map to slice
-		var filteredPkgs []*models.Package
-		for _, pkg := range packageMap {
-			filteredPkgs = append(filteredPkgs, pkg)
-		}
-
-		// Update manifest with filtered packages
-		manifest.Packages = filteredPkgs
+		manifest.Packages = filterDuplicatePackages(manifest.Packages)
```
abhisek, line 75: If we filter these out, then visibility is compromised. One of the goals of `vet` is to build as complete inventory as possible. Whats the rationale for this filtering?

abhisek (approved):

PR #760, opened 2026-07-29
fix(analyzer): compare hostname when matching npm trusted registry URLs

### LINKED ISSUE #759: npm lockfile poisoning: trusted registry URLs with an explicit port never match
### What happened

The npm lockfile poisoning analyzer reports **every** package in `package-lock.json` as "resolved to an untrusted host" when a trusted registry URL in the config includes an explicit port.

In `npmIsTrustedSource` (`pkg/analyzer/lfp_npm.go`), the resolved URL's host is taken with `Hostname()` (port stripped), but the trusted URL is compared using `Host` (port included):

```go
host := parsedUrl.Hostname()                              // "registry.internal"
...
if !strings.EqualFold(parsedTrustedUrl.Host, host) {      // "registry.internal:8443"
    continue
}

if parsedTrustedUrl.Port() != "" && parsedTrustedUrl.Port() != port {
    continue
}
```

The two values can never be equal when the trusted URL has a port, so the loop always `continue`s and the dedicated port comparison immediately below is unreachable — meaning a genuine port mismatch is not enforced either.

### Reproduce

```go
npmIsTrustedSource("https://registry.example.org:8443/a/b/-/c.tgz",
    []string{"https://registry.example.org:8443"})
// got: false, want: true
```

| `trusted_registry_urls` | resolved URL | actual | expected |
| --- | --- | --- | --- |
| `https://registry.example.org:8443` | `https://registry.example.org:8443/a/b/-/c.tgz` | `false` | `true` |
| `https://registry.example.org:8443` | `https://registry.example.org:9443/a/b/-/c.tgz` | `false` | `false` (correct, but by accident — the port check never runs) |

### Expected behaviour

A resolved URL whose scheme, host, port and base path match a configured trusted registry URL should be treated as trusted, regardless of whether that URL carries an explicit port.

### Impact

Anyone whose `trusted_registry_urls` points at a self-hosted registry on a non-default port (Artifactory, Nexus, Verdaccio, an internal proxy) gets one `LockfilePoisoning` threat per dependency. The signal is drowned in false positives, so real lockfile poisoning becomes easy to miss.

### Environment

- `vet` `main` (commit c41b8e2)
- Go 1.26.5, darwin/arm64

yuvrajnode:
Fixes #759

`npmIsTrustedSource` compared the trusted URL's `Host` (port included) against the resolved URL's `Hostname()` (port stripped), so any trusted registry URL configured with an explicit port could never match — and the dedicated port comparison below it was unreachable.

This compares `Hostname()` on both sides and lets the existing port check do its job.

**Tests:** adds cases for matching port, mismatched port, trusted-has-port/source-doesn't, and port + base path. `go test ./pkg/analyzer/...` passes and `go vet ./...` is clean.


abhisek (approved):

yuvrajnode:
Looks like the run-e2e / run-test workflows are waiting on maintainer approval. Happy to wait, just flagging in case it slipped by.     @abhisek @KunalSin9h @sudhanshutech

Issue #739, opened 2026-05-27
[Early Signal] Vet scan issues with pypi packages.


KunalSin9h:
pypi package: `epochly@0.6.3` is suspicous

`vet inspect malware --purl PURl` correctly findings suspicious. 

<img width="737" height="311" alt="Image" src="https://github.com/user-attachments/assets/b5a9290e-6ef9-4c06-b820-13a50df9d427" />

`vet scan --purl PURL` is not able to identify it. (+ `vet scan -D .` with `requirements.txt` is also same issue). 

<img width="646" height="676" alt="Image" src="https://github.com/user-attachments/assets/173a2ec3-652b-4c3d-85aa-97f6137b192d" />

KunalSin9h:
@mrswastik-robot go ahead!!

yuvrajnode:
I spent some time narrowing this down. @mrswastik-robot  are you still working on it? Not looking to step on your work, just sharing what I ruled out in case it saves you time.

Two hypotheses I was able to eliminate:

**1. The malware lookup was silently disabled.** Not the case `--malware-query` defaults to `true` (`scan.go`), so the query enricher is registered on a plain `vet scan` run.

**2. The PyPI purl maps to the wrong ecosystem, so the lookup misses.** Also not the case. Both `pkg:/pypi/epochly@0.6.3` and the canonical `pkg:pypi/epochly@0.6.3` resolve to `ECOSYSTEM_PYPI` with name `epochly` / version `0.6.3` through `purlReader`, so the `QueryPackageAnalysis` request that `vet scan` sends is well-formed.

That leaves the difference in what the two commands actually do. `vet scan` only *queries* existing analysis records via `QueryPackageAnalysis` (`pkg/scanner/enrich_malware_query.go`), whereas `vet inspect malware` at v1.17.2 *submitted* the package and waited for a fresh verdict  visible in the first screenshot ("Submitted package for malware analysis with ID: ..."). So the question is whether a `SUSPICIOUS` / `MEDIUM`-confidence automated verdict ends up in the queryable threat intelligence database at all. I can't check that from the client side.

**Question for maintainers:** given #752 removed on-demand analysis from `vet inspect malware`, is query-only the intended behaviour for both commands now? If so, is the expectation that suspicious-but-unverified automated results become queryable, i.e. is the fix here server-side rather than in this repo?

**One client-side issue I did find regardless of the above.** When the malware lookup fails for any reason, the error is only sent to the log file:

```go
[...]

PR #761, opened 2026-07-29
fix(reporter): surface malicious package lookups that never completed

yuvrajnode:
Partially addresses #739. **Not a close** — see the caveat at the bottom.

## Problem

When a malicious package lookup fails, the error only reaches the log file. `packageEnrichWorkQueueHandler` logs and swallows it:

```go
err := enricher.Enrich(item, s.packageDependencyHandler(pm, item, q))
if err != nil {
    logger.Errorf("Enricher %s failed with %v", enricher.Name(), err)
}
```

(`pkg/scanner/scanner.go`)

The summary then prints:

```
** 0/1 libraries were scanned for malware
```

which is indistinguishable from "we looked up every package and they all came back clean". Because logs are redirected to a file by default, the user gets no signal that the check never completed. In #739 this reads as vet being unable to identify a package that `vet inspect malware` flags as suspicious.

## Change

Track failed lookups on the package manifest and surface the count at the end of the summary and markdown summary reports. This mirrors the existing mechanism for quota errors (`IncrementMalwareAnalysisQuotaError` → `malwareAnalysisQuotaLimitErrorCount` → rendered after the findings), so the two error classes stay distinct: quota-denied lookups keep their existing pricing message, everything else (transport failure, timeout, package absent from the threat intelligence database) counts as a failed lookup.

New output when lookups fail:

```
3 malicious package lookups did not complete. These packages were not
checked for malware. Re-run with `-l -` to see the errors.
```

This does **not** change which packages get flagged. It makes an incomplete malware check visible instead of silent.

## Tests

- `pkg/scanner`: extends the enricher table test to assert the failure is recorded on the manifest. Verified it fails without the fix (`expected: 1, actual: 0`).
- `pkg/reporter`: message rendering test alongside the existing quota one.
- `pkg/models`: counter increments, and stays independent of the quota counter.

`go test ./pkg/models/... ./pkg/reporter/... ./pkg/scanner/...` passes; `go vet ./...` and `gofmt` are clean.

## Caveat on #739

This does not make `vet scan` report `epochly@0.6.3` as suspicious. I ruled out two candidate root causes for that:

- `--malware-query` defaults to `true`, so the query enricher does run on a plain `vet scan`.
- `pkg:/pypi/epochly@0.6.3` resolves correctly to `ECOSYSTEM_PYPI` / `epochly` / `0.6.3` through `purlReader`, so the request is well-formed.

The remaining difference is that `vet scan` only queries existing analysis records, while `vet inspect malware` at v1.17.2 submitted the package and waited for a fresh verdict. Whether a `SUSPICIOUS` / `MEDIUM`-confidence automated result should be queryable is a server-side question I can't verify from the client. Happy to follow up if maintainers can confirm the intended behaviour — I left the details in a comment on #739.

### DIFF
--- a/pkg/models/manifest_test.go
+++ b/pkg/models/manifest_test.go
@@ -113,3 +113,17 @@ func TestContainerImageManifestAddPackages(t *testing.T) {
 	assert.Equal(t, manifest.Id(), pkg1.Manifest.Id())
 	assert.Equal(t, manifest.Id(), pkg2.Manifest.Id())
 }
+
+func TestMalwareAnalysisLookupErrorCount(t *testing.T) {
+	pm := &PackageManifest{}
+
+	assert.Equal(t, 0, pm.GetMalwareAnalysisLookupErrorCount())
+
+	pm.IncrementMalwareAnalysisLookupError()
+	pm.IncrementMalwareAnalysisLookupError()
+
+	assert.Equal(t, 2, pm.GetMalwareAnalysisLookupErrorCount())
+
+	// Lookup errors and quota errors are tracked independently
+	assert.Equal(t, 0, pm.GetMalwareAnalysisQuotaErrorCount())
+}
--- a/pkg/models/models.go
+++ b/pkg/models/models.go
@@ -123,6 +123,11 @@ type PackageManifest struct {
 type internalState struct {
 	// Quota error count for malware analysis api
 	malwareAnalysisQuotaErrorCount int
+
+	// Failed lookup count for malware analysis api. This covers any error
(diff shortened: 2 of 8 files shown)


abhisek:
This is not a feature aligned with the direction of `vet`. It does not proactively scan malicious packages, instead it look a trusted source (SafeDep + OSV) for malicious packages which it correlates with its inventory for reporting and policy decision. Packages that are not found in the database are "unknown". It should not be considered as an error, since `vet` does not actually performs proactive scanning.

Issue #769, opened 2026-08-24
Package version is not resolved: purl copied verbatim, operator dropped, 0.0.0 invented


arunanshub:
`PackageDetails` splits the two facts by design (`pkg/parser/custom/packagefile/packagefile.go:37-40`):

```go
// Version extracted. It can be min, max or exact. It can be empty or exact version string
Version string
// Specs specific version string with operators
VersionExpr string
```

Only `Version` reaches a report and the Control Tower sync. Three parse paths fill it, and none of them resolve a version. Two lie instead.

## 1. The purl path copies the version verbatim

`pkg/common/purl/purl.go:40` assigns `instance.Version` raw. A purl version segment is free text, so `latest` and an operator both survive. CycloneDX (`pkg/parser/cyclonedx.go:134`) and SPDX with a purl reference (`spdx.go:84`) both take this path.

```console
$ cat > bom.json <<'JSON'
{
  "bomFormat": "CycloneDX", "specVersion": "1.5", "version": 1,
  "metadata": { "component": { "type": "application", "name": "demo", "bom-ref": "root" } },
  "components": [
    { "type": "library", "name": "lodash", "version": "latest",
      "purl": "pkg:npm/lodash@latest", "bom-ref": "pkg:npm/lodash@latest" },
    { "type": "library", "name": "express", "version": "^4.0.0",
      "purl": "pkg:npm/express@%5E4.0.0", "bom-ref": "pkg:npm/express@%5E4.0.0" }
  ],
  "dependencies": [
    { "ref": "root", "dependsOn": ["pkg:npm/lodash@latest", "pkg:npm/express@%5E4.0.0"] }
  ]
}
JSON

$ vet scan -L bom.json --lockfile-as bom-cyclonedx \
    --enrich=false --malware-query=false --report-json out.json
```

`out.json`:

```
CycloneDxSBOM  lodash     "latest"
CycloneDxSBOM  express    "^4.0.0"
```

## 2. The text path drops the operator and keeps the operand

`pkg/parser/custom/py/setuppy.go:65-70` cuts the line at the operator and keeps the right side.

```console
$ cat > setup.py <<'PY'
from setuptools import setup
setup(name="demo", install_requires=["requests>=2.28", "django~=4.2"])
PY

$ vet scan -L setup.py --lockfile-as setup.py \
    --enrich=false --malware-query=false --report-json out.json
```

`out.json`:

```
PyPI  requests  "2.28"
PyPI  django    "4.2"
```

[...]

Issue #343, opened 2025-02-05
Duplicate & incorrect version resolution when optional dependencies are specified in requirements.txt


OmkarPh:
Example requirements.txt 
```txt
bleach==3.1.2
bleach[css]
 ```
Expected - Single bleach package, with version `3.1.2`
Actual - Multiple entries for bleach, one with correct version `3.1.2` and another with `0.0.0`

![Image](https://github.com/user-attachments/assets/50659789-fc91-493e-aba0-9748d51fd986)

abhisek:
@Sahilb315 Adding this for you to look at after you are done with your existing stuff.

Sahilb315:
What parser are we using for parsing the requirements.txt file? I am able to get the dependency graph but its not specific to the requirements.txt parsing & can affect other parsers

abhisek:
@Sahilb315 Have a look at: https://github.com/safedep/vet/blob/main/pkg/parser/parser.go#L159

PR #770, opened 2026-08-24
fix(parser): stop inventing a package version

arunanshub:
Narrows #769. See the [scope comment](https://github.com/safedep/vet/issues/769#issuecomment-5406011376) — the maintainer rejected that issue's premise for the text and purl paths, so this PR fixes only the part that loses packages.

## The rule

vet chooses coverage over accuracy. The package name carries most of the security signal: a malicious package whose every version is malicious must still be reported from `flask>=1.0`, and `flassk>=1.0` is an attack the name alone reveals. So **no path may drop a package because its version is imperfect.**

`filterDuplicatePackages` drops any package whose version is `""` or `0.0.0`. That single fact decides everything below.

## What changed

New `pkg/common/utils/version`. `BestEffort` returns the version an expression names, or the lowest one it bounds:

| Input | Result |
| --- | --- |
| `4.17.21`, `1.0.0rc1`, `v1.2.3`, `1.1.0-SNAPSHOT`, `main`, `latest`, `2.x` | unchanged |
| `>=2.28`, `^4.0.0`, `~=4.2`, `!=2.0`, `<2` | `2.28`, `4.0.0`, `4.2`, `2.0`, `2` |
| `>= 1.5.6,< 1.5.7 \|\| > 1.5.7`, `[1.0,2.0)` | `1.5.6`, `1.0` |
| `>=`, `==`, `` | empty — the expression names no version |

**SPDX** (`spdx.go`) drops the `0.0.0` default. The old regex `^\s*([<>=!~]+)?\s*([0-9]+\.[0-9]+(\.[0-9]+)?)?\s*$` rewrote every version it did not match, and the package was then dropped. This is the coverage bug in #769.

**setup.py** (`setuppy.go`) keeps flooring, and now floors `<`, `>` and `!=` as well. Those three were previously lost: `urllib3<2` produced the name `urllib3<2` with version `0.0.0`, so the package vanished.

**purl** (`purl.go`, `packagefile.go`) — reverted, byte-identical to `main`. A purl version passes through verbatim, so a CycloneDX component with a range keeps its package and `vet scan --purl` keeps working.

## Coverage: package counts on the fixtures in this repo

| Fixture | `main` | this PR |
| --- | --- | --- |
| `janusgraph_oss_2dc3a123d9.json` (SPDX) | 200 | **215** |
| `osv-scanner_google_3cab6.json` (SPDX) | 145 | **162** |
| `requests_psf_2ee5b0b01.json` (SPDX) | 5 | **17** |
| `bom-maven.json`, `bom-dropwizard`, `bom-juiceshop` (CycloneDX) | 11 / 167 / 838 | 11 / 167 / 838 |
| `setup2_parser1.py` | 2 | 2 |

**+44 packages, zero losses.** The recovered ones are real versions the regex destroyed: `1.1.0-SNAPSHOT`, `9.4.53.v20231009`, `32.1.3-jre`, `4.2.0-incubating`, Go pseudo-versions, git commits, and every GitHub Actions ref, which is a bare `2` or `main`.

## Repro from #769

```
############ main ############                ############ this PR ############
--- bom.json (bom-cyclonedx)                  --- bom.json (bom-cyclonedx)
    express   '^4.0.0'                            express   '^4.0.0'
    lodash    'latest'                            lodash    'latest'
    react     '17.0.2'                            react     '17.0.2'
--- setup.py                                  --- setup.py
    django    '4.2'                               django    '4.2'
    flask     '3.0.0'                             flask     '3.0.0'
    flassk    '1.0'                               flassk    '1.0'
    requests  '2.28'                              requests  '2.28'
                                                  urllib3   '2'        <- was dropped
--- spdx.json (bom-spdx)                      --- spdx.json (bom-spdx)
    acme/delta '2.28'                             acme/beta        'v1.2.3'
    acme/zeta  '4.17.21'                          acme/delta       '2.28'
                                                  acme/gamma       '1.0.0-beta.1+build.7'
                                                  acme/zeta        '4.17.21'
                                                  actions:checkout '4'
                                                  alpha            '1.0.0rc1'
```

## Tests

`TestBestEffort` carries the version strings the repo's own SPDX, CycloneDX and lockfile fixtures hold, rather than invented ones: a Maven qualifier, a four-component release, a Go pseudo-version, a git commit, a git ref, a `git describe` output, and the range expressions those fixtures put in a version field.

`TestBestEffortKeepsAVersionWhereverOneExists` and `FuzzBestEffort` both fail if an expression holding a version ever yields an empty one again. That is the regression this PR had and the review caught.

`FuzzBestEffort` seeds from the table and guards four invariants: `BestEffort` never panics, it never returns a value the input does not hold, a version resolves to itself, and an expression holding a digit never yields empty.

The idempotence invariant earned its place — it found two real bugs in 2c96caf. `BestEffort(",>=1.0")` returned `>=1.0`, operator and all, because the code stripped operators before it split clauses, so an operator opening a later clause survived. `BestEffort("!\n0")` returned `"\n0"`, because a newline broke no clause but `TrimSpace` removed it on the next pass. Splitting on any whitespace as well as `,` and `|`, then stripping per clause, makes a returned version settle in one pass by construction. 13.4M executions are now clean.

Mutation results, reported as measured: never-invents (a `0.0.0` fallback), idempotence (stripping one operator per clause), and digit-implies-non-empty (rejecting bounds — the exact shape this review rejected) each fail a test. No-panic is **not** kill-capable any more: the implementation holds no index arithmetic, only `FieldsFunc` and `TrimLeft`, so it guards a future rewrite rather than today's code.

## Residual risk

- `NOASSERTION`, `NONE` and `*` now pass through as the version instead of becoming `0.0.0`. The package survives, which the rule wants, but the version string is junk. Reporting a name with no version at all needs a change to `filterDuplicatePackages`, which also guards #343 — separate PR.
- `!=2.0` floors to `2.0`, the one version the spec excludes. It keeps the package, where `main` dropped it.
- A floored version is approximate by design. An exact-version consumer, such as the malware query or a Control Tower identity row, reads it as exact.

## Follow-ups, not in this PR

- `cyclonedx.go:139` keys `bomRefMap` by the purl string while `dependencies[].ref` carries `bom-ref`. When they differ, every relation logs `not found in bomRefMap` and the graph is empty.
- `spdx.go` `attempParsePackageName` gives a package the *name* `0.0.0` when the name is empty.
- `filterDuplicatePackages` logs a drop at debug level, so a lost package leaves no visible trace.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/parser/custom/py/setuppy_test.go
+++ b/pkg/parser/custom/py/setuppy_test.go
@@ -29,7 +29,7 @@ func TestParseRequirementsFileLine(t *testing.T) {
 			input: "flask>=1.0",
 			expected: lockfile.PackageDetails{
 				Name:      "flask",
-				Version:   "",
+				Version:   "1.0",
 				Ecosystem: lockfile.PipEcosystem,
 				CompareAs: lockfile.PipEcosystem,
 			},
@@ -39,7 +39,7 @@ func TestParseRequirementsFileLine(t *testing.T) {
 			input: "numpy~=1.20",
 			expected: lockfile.PackageDetails{
 				Name:      "numpy",
-				Version:   "",
+				Version:   "1.20",
 				Ecosystem: lockfile.PipEcosystem,
 				CompareAs: lockfile.PipEcosystem,
 			},
@@ -49,7 +49,7 @@ func TestParseRequirementsFileLine(t *testing.T) {
 			input: "django!=2.0",
 			expected: lockfile.PackageDetails{
 				Name:      "django",
-				Version:   "",
+				Version:   "2.0",
 				Ecosystem: lockfile.PipEcosystem,
 				CompareAs: lockfile.PipEcosystem,
 			},
@@ -69,7 +69,7 @@ func TestParseRequirementsFileLine(t *testing.T) {
 			input: "urllib3<2",
 			expected: lockfile.PackageDetails{
 				Name:      "urllib3",
-				Version:   "",
+				Version:   "2",
 				Ecosystem: lockfile.PipEcosystem,
 				CompareAs: lockfile.PipEcosystem,
 			},
@@ -148,7 +148,7 @@ func TestParseSetuppy(t *testing.T) {
 				},
 				{
 					Name:      "google-cloud-pubsub",
-					Version:   "",
+					Version:   "2.0",
 					Ecosystem: lockfile.PipEcosystem,
 					CompareAs: lockfile.PipEcosystem,
 				},
--- a/pkg/parser/custom/py/setuppy_test.go
+++ b/pkg/parser/custom/py/setuppy_test.go
@@ -29,7 +29,7 @@ func TestParseRequirementsFileLine(t *testing.T) {
 			input: "flask>=1.0",
 			expected: lockfile.PackageDetails{
 				Name:      "flask",
-				Version:   "1.0",
+				Version:   "",
 				Ecosystem: lockfile.PipEcosystem,
 				CompareAs: lockfile.PipEcosystem,
 			},
@@ -39,7 +39,7 @@ func TestParseRequirementsFileLine(t *testing.T) {
 			input: "numpy~=1.20",
 			expected: lockfile.PackageDetails{
 				Name:      "numpy",
-				Version:   "1.20",
+				Version:   "",
 				Ecosystem: lockfile.PipEcosystem,
 				CompareAs: lockfile.PipEcosystem,
 			},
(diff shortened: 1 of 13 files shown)


pkg/parser/custom/py/setuppy_test.go
```diff
@@ -29,7 +29,7 @@ func TestParseRequirementsFileLine(t *testing.T) {
 			input: "flask>=1.0",
 			expected: lockfile.PackageDetails{
 				Name:      "flask",
-				Version:   "1.0",
```
abhisek, line 32: This is not acceptable from vet's design goals. Here is why:

- vet scans a project or in future a full file system
- vet finds a known malicious package (all versions are malicious) in requirements.txt or equivalent

Today vet will still report it and will flag it as malicious even if its version resolution is "best effort" and it cannot truly resolve a spec like "flask>=1.0", but in reality, it will detect "flask" and a best effort version "1.0".

This is incredibly useful from security perspective due to:

1. There is a dependency on flask, even though we don't know the exact version
2. Instead of flask, if its flassk, and then its pretty much an attack detection

vet can't miss these cases. It should choose coverage over accuracy.

arunanshub: Agreed, and reverted in 42b6148. Version resolution is best effort again: `flask>=1.0` reports flask at `1.0`, a multi-clause spec floors to its first clause, and no path drops a package for an imperfect version. My earlier shape returned an empty version, and `filterDuplicatePackages` drops a package whose version is empty or `0.0.0`, so it lost the name too, which is the signal that matters.

I also reverted the purl guard, so a CycloneDX component carrying a range keeps its package and passes the version through exactly as before. Say the word if you would rather it floored like the text path.

The SPDX change stays, because it fixes the same class of loss in the other direction: the old regex rewrote every version it did not match to `0.0.0`, so the package was dropped. Against the SPDX fixtures in this repo the reports grow from 200 to 215, 145 to 162, and 5 to 17 packages — `1.1.0-SNAPSHOT`, `v1.2.3`, Go pseudo-versions and every bare GitHub Actions ref were being lost. A test and a fuzz invariant now fail if an expression holding a version ever resolves to empty again.

abhisek (approved):

PR #771, opened 2026-08-25
feat(auth): retry transient failures and time out cloud gRPC calls

### LINKED ISSUE #508: Handle Retries and Timeouts for vet and SafeDep Cloud Connection
We need to implement a generic timeout handling and retry layer for gRPC connection between `vet` and SafeDep Cloud. This should include:

- Timeout handling
- Intermittent error handling, especially which may be triggered during platform scaling events

Ref: https://github.com/safedep/vet-action/issues/112

ChrisJr404:
Fixes #508.

The cloud gRPC clients were dialed with an empty set of options, so there was no default deadline and no per-call retry. If the control plane was briefly unavailable (for example during a scaling event) or a call hung, that surfaced as an immediate failure to the user.

This injects a small set of default dial options in `cloudClientConnection`:

- a default per-call timeout so a stuck RPC eventually errors instead of hanging
- retry of `UNAVAILABLE` with exponential backoff, which covers the transient "no ready backend" window during scaling
- bounded reconnect backoff (`WithConnectParams`) so a client recovers quickly once the backend is healthy again, rather than sitting in the default multi-minute backoff

The happy path is unchanged. Retries only kick in for `UNAVAILABLE`, which means the RPC never reached a serving backend, so this is safe for the non-idempotent calls too.

Both knobs have sensible defaults (4 attempts, 30s timeout) and can be tuned or disabled with `VET_CLOUD_GRPC_MAX_ATTEMPTS` (1 disables retries) and `VET_CLOUD_GRPC_TIMEOUT` (`0s` disables the deadline).

Added tests validate the generated service config against gRPC and cover the env parsing, clamping, and disable paths. `go test ./internal/auth/...`, `go vet`, and `golangci-lint run` are clean.

### DIFF
--- a/internal/auth/grpc.go
+++ b/internal/auth/grpc.go
@@ -1,15 +1,32 @@
 package auth
 
 import (
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"net/url"
 	"os"
+	"strconv"
+	"time"
 
 	drygrpc "github.com/safedep/dry/adapters/grpc"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/backoff"
 
 	"github.com/safedep/vet/pkg/common/logger"
+)
+
+const (
+	cloudGrpcMaxAttemptsEnvKey = "VET_CLOUD_GRPC_MAX_ATTEMPTS"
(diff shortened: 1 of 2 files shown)


abhisek (approved):

abhisek:
Thank @ChrisJr404 for your contribution.

PR #772, opened 2026-08-27
feat(mdm): per-user endpoint scan script for MDM fleets

Sahilb315:
## What

Two changes that together let a fleet scan every user on a shared machine and attribute each user's inventory correctly in SafeDep Cloud:

1. **`scripts/mdm/vet_endpoint_scan.sh`** — one cross-platform (Linux + macOS) script that runs `vet endpoint scan` for every local user and syncs to the cloud. Modeled on the [PMG MDM scripts](https://github.com/safedep/pmg/tree/main/scripts/mdm).
2. **Stamp the OS user on cloud events** — the cloud sink now records `username` + `uid` on every event's `EndpointInvocationContext`, so users on one host are distinguishable.

## Why

Scanning a shared machine as root points `HOME`, per-user config, and credentials at root's world, not the user's. Switching to each user is OS-level orchestration (`sudo -u`) that belongs in a deployment script, not inside `vet`. The one piece that *is* vet's job — reporting which OS user a scan ran as — is the stamping change.

## Script (`scripts/mdm/vet_endpoint_scan.sh`)

- **As root** (typical MDM): scans every local human account, each under its own context via `sudo -u`. Linux `/etc/passwd` (UID ≥ `UID_MIN`, home under `/home`); macOS Directory Services (UID ≥ 500, home under `/Users`).
- **As the logged-in user**: scans just that user, with their full login environment.
- Cloud credentials (`SAFEDEP_API_KEY` / `SAFEDEP_TENANT_ID`) are streamed into each scan over **stdin**, never argv, so they don't leak through `ps` on a multi-user box.
- Extra args are forwarded to `vet endpoint scan` (e.g. `--silent --kind ai-tool`).

## Cloud attribution (`pkg/inventory/sinks/cloud`)

- Resolve the running user via `user.Current()` once in `New`; stamp `username` + `username_uid` onto every `ToolEvent.InvocationContext` in `send()`.
- `hostname` + `machine_id` are identical for every user on a box; the invocation-context fields are what let the cloud separate them.
- A `user.Current()` failure degrades gracefully — context omitted, scan continues.
- **Requires a dependency bump**: the `EndpointInvocationContext.username`/`username_uid` fields landed in the safedep proto after vet's pinned snapshot. Bumped `buf.build/gen/go/safedep/api/{protocolbuffers,grpc}/go` to the `2026-07-21` snapshot (first to carry the fields; keeps protobuf at v1.36.11). This pulls minor transitive bumps: `google.golang.org/grpc` 1.80→1.81, `golang.org/x/oauth2` 0.35→0.36, otel gcp detector 1.41→1.42.

## Tests

- `scripts/mdm/tests/vet_endpoint_scan_test.sh` (runs as a normal user): OS guard, binary resolution, credential pairing, current-user scan with a mock `vet` (asserts creds arrive via env, never argv), local-only mode, Linux enumeration filter. New `mdm.yml` workflow runs it on Linux + macOS + `shellcheck`.
- `TestCloudSink_StampsUserInvocationContext` / `TestCloudSink_OmitsInvocationContextWhenIdentityUnknown` in `sink_test.go`.
- Verified locally: `go build ./...`, `go test ./pkg/inventory/...`, `gofmt`, `go vet` all clean; MDM script tests green on bash 3.2 and 5.x, and end-to-end with the real `vet` binary.

## Notes

- Under `sudo -u -H`, CLI-tool discovery reads the sanitized `PATH`; running as the logged-in user (MDM user-context payload) gives full CLI-tool fidelity. Documented in the script README.
- The `username`/`uid` stamping is generic ambient context; it could later move into `dry`'s `endpointsync.NewEvent()` so PMG and gryph inherit it too. Kept in vet here to ship without a `dry` release.

Draft: opening for review of the approach, the enumeration/credential handling, and the dependency bump before finalizing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


scripts/mdm/vet_endpoint_scan.sh
```diff
@@ -0,0 +1,206 @@
...
+  done
+  return 1
+}
+
+# Emit "user<TAB>uid<TAB>home" for every local human account on Linux, parsed
+# from the passwd database. Filters to UID >= UID_MIN (default 1000) with a real
+# home under the required prefix. Split out from each_target_user so it is
+# testable without root.
+enumerate_linux_users() {
+  local passwd="${VET_MDM_PASSWD_FILE:-/etc/passwd}"
```
abhisek: Linux enumeration reads only `/etc/passwd`, but macOS enumeration reads Directory Services (`dscl`). On an MDM-managed Linux host that gets its users from LDAP/SSSD/AD, `/etc/passwd` does not list the directory users. So this path silently skips the exact fleet population the script targets, while the macOS path finds them.

Use `getent passwd` instead. It reads NSS, so it returns local and directory users, and it still returns local-only accounts on a standalone host:

```sh
getent passwd | while IFS=: read -r user _ uid _ _ home _; do ...
```

Keep a test seam for the source (for example a command shim), so the tests can still run as a normal user.

If you skip directory users on purpose, please add a comment that says so.

---
_Generated by [Claude Code](https://claude.ai/code)_

scripts/mdm/vet_endpoint_scan.sh
```diff
@@ -0,0 +1,206 @@
...
+  local home_prefix="${VET_MDM_HOME_PREFIX:-/home}"
+  local uid_min
+  uid_min=$(awk '/^UID_MIN/{print $2}' /etc/login.defs 2>/dev/null || true)
+  [[ "$uid_min" =~ ^[0-9]+$ ]] || uid_min=1000
+  local user uid home
+  while IFS=: read -r user _ uid _ _ home _; do
```
abhisek: This filter reads the login shell (the 7th field, dropped as the trailing `_`) but does not check it. A service account with a UID at or above `UID_MIN` and a home under `/home` still gets a full scan, even when its shell is `/usr/sbin/nologin` or `/bin/false`.

Skip accounts that cannot log in. For example:

```sh
while IFS=: read -r user _ uid _ _ home shell; do
  ...
  case "$shell" in */nologin|*/false) continue ;; esac
  ...
```

Low severity. It only wastes work today. It also keeps the Linux and macOS user sets closer.

---
_Generated by [Claude Code](https://claude.ai/code)_

abhisek (commented):
Two notes on the Linux user enumeration in `vet_endpoint_scan.sh`. Both sit in the same function. Neither blocks merge, but the first one affects whether the script finds the right users on a directory-backed fleet.

---
_Generated by [Claude Code](https://claude.ai/code)_

abhisek (approved):

PR #773, opened 2026-08-31
feat(parser): parse SPDX documents as a dependency graph

Alokxk:
Closes #276

Replaces `pkg/parser/custom/sbom/spdx` with a graph parser at `pkg/parser/spdx.go`, modelled on `pkg/parser/cyclonedx.go`. The graph anchors on the elements the document describes; `DEPENDS_ON` and `DEPENDENCY_OF` are edges. A package resolves from its PURL, or from its name when the document carries none.

Package counts on the three GitHub Dependency Graph fixtures are unchanged — 22, 162, 300 — and every package is now a graph root

- A document with no dependency relationship keeps its packages and reports no graph. SPDX does not require one.
- `bom-spdx` is no longer experimental. `TestListParser` 24 -> 25.

Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added SPDX SBOM support through dependency graph parsing.
  * Recognizes packages from package URLs or package names, including ecosystem and group information.
  * Resolves SPDX dependency relationships in either direction and identifies direct dependencies.
  * Handles unsupported packages, malformed package URLs, incomplete metadata, and documents without valid relationships gracefully.
  * Excludes the document’s described package from dependency nodes and reports graph availability only when valid dependency connections exist.

PR #779, opened 2026-09-15
fix(parser): say when CycloneDX components were skipped

arpitjain099:
`cdxExtractPackageFromComponent` fails for any PURL type that is not in `PurlTypeToEcosystem`, and `parseSbomCycloneDxAsGraph` logs it and moves on:

```go
ref, pkg, err := cdxExtractPackageFromComponent(component)
if err != nil {
    logger.Errorf("Failed to extract package from component %v: %v", component, err)
    continue
}
```

`main.go:198` wires the logger to `io.Discard` unless `--log` is given, so on a normal run that line goes nowhere. A four component SBOM, one npm package plus an apk, a deb and a hex one:

```
$ vet scan --lockfiles sbom.cdx.json --lockfile-as bom-cyclonedx
  ** across 1 libraries in 1 manifest(s)
 No risky libraries identified
$ echo $?
0
```

Three quarters of the SBOM is gone and nothing says so. The information exists the whole time:

```
$ vet scan --lockfiles sbom.cdx.json --lockfile-as bom-cyclonedx --log -
level=error msg="... failed to map PURL type:apk to known ecosystem"
level=error msg="... failed to map PURL type:deb to known ecosystem"
level=error msg="... failed to map PURL type:hex to known ecosystem"
```

The types worth leading with are apk, deb and rpm rather than hex, because those are the ones vet itself writes. `Package.GetPackageUrl` renders the distro ecosystems found during container scanning through `distroEcosystemPurls`, so an Alpine image produces `pkg:apk/alpine/busybox@1.36.1-r5?distro=v3.20`, and `PurlTypeToEcosystem` has no entry for `apk`, `deb` or `rpm`. vet cannot read back an SBOM it produced for a container image without dropping every OS package in it, silently.

### The change

Tally the skipped components and print one line to stderr when there were any:

```
warning: sbom.cdx.json: skipped 3 of 4 component(s), no ecosystem is mapped for apk (1), deb (1), hex (1). They were not scanned and are not counted in the results below.
```

Nothing new is scanned and no component is resolved that was not resolved before. The run stops implying it looked at the whole SBOM. I kept it to that rather than extending the ecosystem map, since adding a mapping without backend support for that ecosystem would trade this silent gap for a different one, and which of apk, deb and rpm vet wants to support is your call.

stderr rather than the logger because the logger is discarded by default, and `pkg/reporter/skill_reporter.go` already writes user-facing output to `os.Stderr`, so the result stays out of anything piped from stdout.

Scope: CycloneDX only. The SPDX parser shares `PurlTypeToEcosystem` through `pkg/common/utils/sbom`, but it has a `parsePackageFromPackageDetails` fallback, so the behaviour there is different and I have not measured it. #773 is also rewriting that path.

### Verification

Two tests in `pkg/parser/cyclonedx_test.go`, both capturing stderr around the real parser: the four component SBOM above is expected to yield one package and a warning naming `3 of 4` and each of the three types; an SBOM whose components all map is expected to yield two packages and no output at all. The first fails on `main`, the second passes either way and is there so a change that starts warning on healthy input is caught.

`go test ./pkg/parser/... ./pkg/common/...` is green, and `go vet` and `gofmt -l` are clean on the touched package.

## Summary by CodeRabbit

- **Improvements**
  - CycloneDX parsing now reports components that cannot be converted into packages.
  - Skipped components are summarized by package URL type, including an `unknown` category when the type cannot be determined.
  - Warnings include deterministic counts and totals, while detailed extraction failures remain available in logs.
  - Successfully recognized components continue to be included in the resulting manifest, while unresolved components are excluded.


abhisek:
@KunalSin9h Please help support this with review and testing.

KunalSin9h:
@abhisek sure will look into it.

